PR #10, opened 2022-09-19
Update README.md - Super Minor Typo Fix

2140data:
I believe author intends to use "prepend" not "prepent" on Line 63. Small change, big meaning.


callebtc:
Thank you!

PR #12, opened 2022-09-23
Notation/alice to bob

callebtc:


callebtc:
Change notation to: 
`Alice` is user, `Bob` is mint
`hash_to_curve` to `hash_to_point`

PR #16, opened 2022-09-28
Update requirements.txt

SatoshiNakamotoBitcoin:


callebtc:
Hi, can you briefly explain the reason for this PR? 

You would have to update the library in poetry first and then use the command

```
poetry export -f requirements.txt --without-hashes --output requirements.txt
```

to generate the requirements.txt.

PR #28, opened 2022-10-09
change logger to report msats

trbouma:
On startup logger shows the amount in sats, but it is actually in msats, so off by a factor of a 1000. Should change unit to msats.


callebtc:
Thank you! Would you please remove the `Procfile` from the commit?

I would also suggest to show sats instead of meats and divide the balance by `// 1000`, what do you think?

PR #47, opened 2022-10-23
Add entrypoint shell file for use with Docker Compose

fusion44:
This draft PR adds an entry point file to so this image can be more easily used from within a docker-compose environment.

The `build_env.py` file is run when the Docker container is started. Its sole purpose is to create a LNbits wallet and fetch the admin key. It'll populate the `.env` file with the fetched data. *This file should probably be in a subfolder or somewhere.*

Example for legend-regtest: 
https://github.com/fusion44/legend-regtest-enviroment/blob/main/docker-compose.yml
(use make docker-image before starting the evironment)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,2 @@
+docker-image:
+	docker image build -t cashu:0.4.2 .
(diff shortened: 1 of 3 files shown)


callebtc:
Is this still relevant @fusion44?

PR #53, opened 2022-11-04
Initial pass at the wallet api

tim-habitat:
Let me know your thoughts! Happy to change stuff around

I added a couple of things that could be used in the broader app but left it in the `wallet` for now (such as using `pydantic` for settings management)
There would need to have a couple of tests created as well - though wasn't sure how best to spin up a local mint. if you think it might be worth it, we could create a docker container with the mint / mint server to fake it.


cashu/wallet/__init__.py
```diff
@@ -1,3 +1,5 @@
 import sys
+import pkg_resources
 
 sys.tracebacklimit = None  # type: ignore
+__version__ = pkg_resources.get_distribution("cashu").version
```
tim-habitat, line 5: this though requires that the package `cashu` is installed, ie a `poetry install --no-root` will break it

cashu/wallet/api.py
```diff
@@ -0,0 +1,360 @@
...
+                    reserved_date=reserved_date,
+                    id=key,
+                    mint_token_with_secret=SecretStr(token_hidden_secret),
+                    mint_token=mint_token,
+                )
+            )
+    return list_of_pending
+
+
+@app.get("/generate_lock")
```
callebtc, line 294:
```suggestion
@app.put("/lock")
```

cashu/wallet/api.py
```diff
@@ -0,0 +1,360 @@
...
+    all_tokens: bool = Query(False, description="Burn all spent tokens"),
+    force: bool = Query(False, description="Force check on all token"),
+):
+    """Burn spent tokens."""
+    logger.info(f"Wallet status: {wallet.wallet.status()}")
+    if all_tokens:
+        logger.info("Burn all spent tokens")
+        proofs = get_reserved_proofs(wallet.wallet.db)
+    elif force:
+        # chekc all proofs in db
```
callebtc, line 247:
```suggestion
        # check all proofs in db
```

cashu/core/db.py
```diff
@@ -127,7 +127,7 @@ def __init__(self, db_name: str, db_location: str):
                 print(f"Creating database directory: {self.db_location}")
```
callebtc, line 127: Please revert these changes (not relevant for this PR)

cashu/wallet/__api__.pi.py
```diff
@@ -0,0 +1,6 @@
+from cashu.wallet.api import app as wallet_api
```
callebtc, line 1: This is a weird file name. Please revert and add to `__init__.py`?

cashu/wallet/__init__.py
callebtc: I don't get this. Why is this here?

cashu/wallet/settings.py
```diff
@@ -0,0 +1,49 @@
+from pydantic import BaseSettings, AnyUrl, Field, root_validator, SecretStr
```
callebtc, line 1: Unrelated to PR?

pyproject.toml
```diff
@@ -42,4 +47,10 @@ build-backend = "poetry.core.masonry.api"
...
 cashu = "cashu.wallet.cli:cli"
-wallet-test = "tests.test_wallet:test"
\ No newline at end of file
+wallet-test = "tests.test_wallet:test"
+
+[tool.coverage.run]
+source = ["cashu"]
+
+[tool.black]
+line-length = 120
```
callebtc, line 56: Isn't this the default setting?

callebtc:
Thank you for this great PR! There are a couple of unrelated changes in here that are out of the scope of this PR.

Please describe the API in the top comment. I think the design of the API is not optimal yet. Some requests should be PUT's and also the naming of the endpoints is not optimal (e.g. they include underscores, etc.).

I will try to review this asap.

cashu/wallet/api.py
```diff
@@ -0,0 +1,360 @@
...
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Error creating invoice.",
+        )
+    return {"invoice": invoice, "amount": amount}
+
+
+@app.get("/lightning/receive")
+async def receive(
+    amount: int = Query(..., gt=0, description="Amount to receive in satoshis"),
+    invoice_hash: str = Query(..., description="Hash of the invoice to be paid"),
```
callebtc, line 131: these are not standard names, should be called `description_hash` and also needs `unhashed_description` to work with CLN.

cashu/wallet/api.py
```diff
@@ -0,0 +1,360 @@
...
+        logger.info(f"Invoice {invoice.pr} - invoice hash {invoice.hash}")
+    else:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Error creating invoice.",
+        )
+    return {"invoice": invoice, "amount": amount}
+
+
+@app.get("/lightning/receive")
```
callebtc, line 128:
```suggestion
@app.put("/lightning/invoice")
```

cashu/wallet/api.py
```diff
@@ -0,0 +1,360 @@
...
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Insufficient funds.",
+        )
+    _, send_proofs = await wallet.wallet.split_to_send(wallet.wallet.proofs, amount)
+    await wallet.wallet.pay_lightning(send_proofs, invoice)
+    logger.info(f"Wallet status: {wallet.wallet.status()}")
+    return {"amount": amount, "fees": fees}
+
+
+@app.get("/lightning/invoice")
```
callebtc, line 105: Difference between `lightning/invoice` and `lightning/receive` is unclear. Please stick to exactly the same names as in the CLI, do not invent new names if not necessary.

cashu/wallet/settings.py
callebtc: Nice use of `BaseSettings` though, I think we should keep it. Does it play along with the rest of the software though?

cashu/wallet/settings.py
```diff
@@ -0,0 +1,49 @@
...
+    tor: bool = Field(description="Use Tor", default=True)
+    socks_host: str | None = Field(description="Socks host", default=None)
+    socks_port: int = Field(description="Socks port", default=9050)
+    lightning: bool = Field(description="Use Lightning", default=True)
+    lightning_fee_percent: float = Field(description="Lightning fee percent", default=1.0, ge=0.0)
+    lightning_reserve_fee_min: float = Field(description="Lightning reserve fee min", default=4000)
+    mint_private_key: SecretStr | None = Field(description="Mint private key", default=None)
+    lnbits_endpoint: AnyUrl | None = Field(description="Lnbits endpoint", default=None)
+    lnbits_key: SecretStr | None = Field(description="Lnbits key", default=None)
+    max_order: int = Field(description="Max order", default=64)
```
callebtc, line 48: Most of this code is a duplication of the rest of this file.

cashu/wallet/__init__.py
tim-habitat: this is "just" for:
```python
>>>from cashu import wallet
>>>wallet.__version__
'0.5.1'
```
I can remove it if you feel it's unnecessary.

cashu/wallet/settings.py
tim-habitat: fair - i'll remove for now. would you find it useful to use this instead of the other `cashu/core/settings.py` ?

pyproject.toml
tim-habitat: I think the default is `88` https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length
I personnally find 88 too short. Another advantage of specifying here, is to let other contributors know what the projects expects. Again, can change if you prefer another length

cashu/core/db.py
tim-habitat: Well, i agree but it is relevant as before i made those changes, the `api` was hanging trying to connect to the mint. I can exactly recall why, i'll try to go around it but i think this was the way.

cashu/wallet/__api__.pi.py
tim-habitat: Do we want to start the `api` as soon as we import the `wallet` ? 
My thought of having it in a separate file, is that you may want to import the wallet without starting the `fastapi` app.
Though i agree, the filename is weird. I'll change to `__api__.py` for now. But definitely open to putting it in `__init__.py` just sharing a few thougths.

cashu/wallet/api.py
tim-habitat: Sorry about inventing new names. 
I added this function so the `invoice` path didn't hang waiting for conf. that the invoice has been paid ie to replace:
```python
            while time.time() < check_until and not paid:
                time.sleep(3)
                try:
                    await wallet.mint(amount, invoice.hash)
                    paid = True
                    print(" Invoice paid.")
                except Exception as e:
                    # TODO: user error codes!
                    if str(e) == "Error: Lightning invoice not paid yet.":
                        print(".", end="", flush=True)
                        continue
```
from the `cli` maybe that's a wrong. I can make it hang.
Else, i can make this a `invoice` "status" path.
What do you think is more suited?

cashu/wallet/api.py
tim-habitat: updated names, and added `unhashed_description` parameter

cashu/wallet/api.py
```diff
@@ -0,0 +1,360 @@
...
+
+    wallet_name: str = "default"
+    wallet: Wallet = Wallet(
+        url=mint_settings.mint_url,
+        db=str(Path(cashu_settings.cashu_dir) / wallet_name),
+    )
+    active_wallet: bool
+
+
+wallet = WalletModel(active_wallet=True)
```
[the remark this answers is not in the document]
tim-habitat: Yep, sure thing!

tim-habitat:
Thanks for the review and feedback!
I made most of the changes you requested. Some was not feasible for the api to keep working from my view but i may be wrong. Happy to (re)refactor if still not good enough.

> Please describe the API in the top comment.

Not sure what you would want as a description there? I think it might be worth expanding the docstrings of the endpoints, as hopefully those will be the one other devs interact and not necessarly the code directly.

> the naming of the endpoints is not optimal (e.g. they include underscores, etc.

The function are snake cased, but the endpoints themselves don't contain underscore i believe. The query parameters do though, can change that if an issue.

Hopefully, you are happy with this PR - I'm keen on helping out on other stuff if you open issues, i can contribute some more! eg moving to using the `pydantic BaseSettings` or helping review other PRs. Anyway, let me know!
Have a nice week-end!

callebtc:
Unfortunately this PR can't be merged in its stage and since no further work has been done on it, I will close for now!

PR #58, opened 2022-11-22
Update README.md

vnprc:
change brew install postgres to postgresql


vnprc:
I opened this PR and noticed PR #53 has an update for the same line of the readme.

PR #62, opened 2022-11-26
Printing invoice as QR-Code on terminal

PrinzMonty:
I added an optional argument to allow the LN address and the token address to be printed as a QR code.

`cashu invoice 100 --qrcode 1`

https://user-images.githubusercontent.com/4109582/204111000-ffef3036-b174-4e8c-b490-bcce91229948.mp4


callebtc:
Hi. Thank you very much for this PR. I'm not sure about the feature though. I think it's a bit too much, considering that the QR codes for ecash are typically too large. For invoices it might make sense but I doubt that most people have the necessary terminal real estate to show them. What do you think?

PS: Tests are failing

PrinzMonty:
Hey!

You're probably right, I just found it a bit annoying when I first tested Cashu that I didn't have a QR code to scan. That's why I built it as an optional feature for myself. But it's possible that it's not really needed.

I haven't programmed much with Python before, so I didn't know about Poetry either. It's possible that I made a mistake there. I'll take a look again.

PR #63, opened 2022-11-30
Fix/db_timestamps

callebtc:
fix timestamps for postgres


Bashy:
I now get a `ImportError: Please make sure that the extension `cashu` follows conventions.` with the master branch

PR #82, opened 2022-12-28
add description_hash

trbouma:
added description hash arguments


callebtc:
Unfortunately this PR can't be merged in its stage and since no further work has been done on it, I will close for now!

PR #83, opened 2022-12-29
Update b_dhke.py

lackrobin:
I was a little bit confused by this comment, I think the mint and client got mixed up

### DIFF
--- a/cashu/core/b_dhke.py
+++ b/cashu/core/b_dhke.py
@@ -3,28 +3,28 @@
 """
 Implementation of https://gist.github.com/RubenSomsen/be7a4760dd4596d06963d67baf140406
 
-Alice (Client):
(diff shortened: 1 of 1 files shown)


phyro:
@lackrobin good catch. The names were flipped around in the code, but not in the comments. The last line in the comment should also be flipped now "If true, C must have originated from Alice" should now be Bob.

lackrobin:
@phyro thanks! i missed that one. Fix is in #6d0ac86

phyro:
hmm, it seems switching names in code might have made the variable names slightly confusing as Alice now uses `B'` (though B' can be seen as a Blinded pub) and Bob uses `A`. But I guess this is now consistent with the code. I'll go ahead and merge this.

PR #131, opened 2023-03-05
Determinstic secrets / ecash restore

callebtc:
This PR introduces the ability to restore the user's ecash balance from a single secret with the help of the mint. Unfinished but backup and restore basics are working. 

Teaser: https://twitter.com/callebtc/status/1620186555993456641

## User testing
To test this feature, please install cashu from the GitHub repository. Then add the testnut mint that already runs this branch. This is the process:

```bash
cd cashu/
git pull && git checkout deterministic_secrets
poetry install && poetry shell

# add the testnut mint to this wallet
echo "MINT_URL=https://testnut.cashu.space" > .env 
```

Your wallet should be ok now. You can now mint tokens, send them to yourself, delete your wallet and restore it:

```bash
# running cashu once will now generate a seed phrase for you
# you can check the seed phrase with and copy it somewhere
cashu info --mnemonic

# now for the fun part
cashu invoice 12345
cashu send 10
cashu receive <token>

# let's delete your wallet
rm -r data/wallet

# let's restore it - enter your seed phrase
cashu restore

# your balance should be restored now
cashu balance
```

## Deterministic secret generation
The wallet generates a private key and has a counter (called `counter` below) that starts at `0`. There are two (previously random) inputs for token generation: `secret` and `blinding_factor` to generate `BlindedMessages`. 

Previously, we used random numbers for these. Here, we generate them deterministically from the `private_key` and a counter that increments for every token we create. 

# Deriving secrets

We produce a standardized BIP32-compatible derivation path for generating `secret` and blinding factor `r` for each token. The path encodes:

- purpose' = `129372'` (UTF-8 for 🥜)
- coin type' = Bitcoin `0'`
- keyset id' = Keyset ID as integer `keyset_id_int'`
- coin counter' = `counter'` <-- this one goes up
- secret or r = `0` or `1`

Thus:
```python
secret_derivation_path = f"m/129372'/0'/{keyset_id_int}'/{counter}'/0"
r_derivation_path = f"m/129372'/0'/{keyset_id_int}'/{counter}'/1"
```

The wallet increments `counter` after every derivation. When the wallet restores its funds, it restarts `counter` at `0` and increments it as long as necessary to restore its balance.

## Quirks
When a wallet sends a `POST /mint` request, it attaches `BlindMessages` to it that will necessarily increment the `counter`. However, if the invoice has not been paid yet, they won't be converted into `Proofs`. Therefore, say, if the wallet would attach 10 `BlindMessages` and tried to request the mint 5 times without success, the `counter` would be incremented by 50 without having any corresponding `Proofs` for the appropriate values. The number of missed `counter` values a wallet might have missed is not easy to predict either. Therefore, a wallet **MUST** not increment `counter` for failed `POST /mint` requests but only for successful ones.

# Backup restore
After a database wipe, we can use the deterministic derivation to re-generate the same `BlindedMessages` that we had generated before. Generally, the user won't remember the state of their last `counter` before the database wipe, so we generate many of these, say `100` per request. 

We can send these 100 `BlindedMessages` to the mint and ask it politely to re-issue the `BlindedSignatures` for them. The mint was so kind to keep these in their database. The mint matches the provided `BlindedMessages` against its database and returns all `BlindedSignatures` it issued for these matches.

The wallet uses the endpoint `POST /restore` and sends an ordinary `PostMintRequest` (which includes the `BlindedMessages` (this is the same object we use for `POST /mint`).

The wallet then receives the object `PostRestoreResponse`:
```python
class PostRestoreResponse(BaseModel):
    outputs: List[BlindedMessage] = []
    promises: List[BlindedSignature] = []
```

The fields `outputs` and promises` only contain positive matches. If we had 10 tokens during the wallet's life time, we will receive 10 matches (and not 100 objects as we sent to the mint). The wallet now needs to take these `outputs` and filter the the correct `secret` and `blinding_factor` from the `100` it generated in the beginning of the restore process. 

# Things to decide:

- [x] How to generate private keys and let user enter private keys? -> Nost private key for now
- [x] How to derive secrets? -> BIP32
- [x] Should I derive `secret` from `r` or generate them independently? -> independently

Todo:
- [x] Verify that failed `POST /mint` requests do not increment the counter
- [x] Restore tokens from multiple mints
- [x] Store derivation path of each `Proof` in the db
- [x] Store private key in database
- [x] Restore tokens for multiple wallets


cashu/mint/migrations.py
```diff
@@ -159,3 +159,20 @@ async def m006_invoices_add_payment_hash(db: Database):
...
+    """
+    await db.execute(
+        f"ALTER TABLE {table_with_schema(db, 'proofs_used')} ADD COLUMN id TEXT"
+    )
+    await db.execute(
+        f"ALTER TABLE {table_with_schema(db, 'proofs_pending')} ADD COLUMN id TEXT"
+    )
+    await db.execute(
+        f"ALTER TABLE {table_with_schema(db, 'promises')} ADD COLUMN id TEXT"
+    )
```
callebtc, line 178: Migration to add `id` column to `proofs_used`, `proofs_pending` and `promises`.

cashu/mint/ledger.py
```diff
@@ -177,7 +177,11 @@ async def _generate_promise(
         C_ = b_dhke.step2_bob(B_, private_key_amount)
         logger.trace(f"crud: _generate_promise storing promise for {amount}")
         await self.crud.store_promise(
-            amount=amount, B_=B_.serialize().hex(), C_=C_.serialize().hex(), db=self.db
+            amount=amount,
+            B_=B_.serialize().hex(),
+            C_=C_.serialize().hex(),
+            id=keyset.id,
```
callebtc, line 183: we now store the "id" for a promise in the wallet db because we need it for restore

cashu/mint/ledger.py
```diff
@@ -932,3 +936,20 @@ async def split(
...
+            for output in outputs:
+                logger.trace(f"looking for promise: {output}")
+                promise = await self.crud.get_promise(
+                    B_=output.B_, db=self.db, conn=conn
+                )
+                if promise is not None:
+                    promises.append(promise)
+                    return_outputs.append(output)
+                    logger.trace(f"promise found: {promise}")
+        return return_outputs, promises
```
callebtc, line 955: Mint: that's all there is

cashu/mint/migrations.py
callebtc, line 178: Store keyset ID in all objects in db

cashu/wallet/cli/cli.py
```diff
@@ -81,31 +108,40 @@ def cli(ctx: Context, host: str, walletname: str):
     ctx.ensure_object(dict)
     ctx.obj["HOST"] = host or settings.mint_url
     ctx.obj["WALLET_NAME"] = walletname
-    wallet = Wallet(
-        ctx.obj["HOST"], os.path.join(settings.cashu_dir, walletname), name=walletname
-    )
+
+    db_path = os.path.join(settings.cashu_dir, walletname)
+    # loop = asyncio.new_event_loop()
```
xphade, line 108: deprecated? also below.

cashu/wallet/cli/cli.py
```diff
@@ -81,31 +108,40 @@ def cli(ctx: Context, host: str, walletname: str):
...
+    # if the command is "restore" we don't want to ask the user for a mnemonic
+    # otherwise it will create a mnemonic and store it in the database
+    if ctx.invoked_subcommand == "restore":
+        wallet = await Wallet.with_db(
+            ctx.obj["HOST"], db_path, name=walletname, skip_private_key=True
+        )
+    else:
+        # # we need to run the migrations before we load the wallet for the first time
+        # # otherwise the wallet will not be able to generate a new private key and store it
+        # loop.run_until_complete(migrate_wallet_db(Database("wallet", db_path)))
```
xphade, line 119: Is this comment also deprecated? Because it looks like the migration is actually not done.

tests/conftest.py
```diff
@@ -94,19 +81,47 @@ async def start_mint_init(ledger):
...
+# def wallet():
+#     wallet = Wallet(
+#         url=settings.mint_host,
+#         db="data/test_wallet",
+#         name="wallet",
+#         private_key="TEST_WALLET_CLI_PRIVATE_KEY_1",
+#     )
+#     asyncio.run(migrate_databases(wallet.db, migrations))
+#     asyncio.run(wallet.load_proofs())
+#     yield wallet
```
xphade, line 127: Why are these here? They seem to be deprecated.

cashu/wallet/wallet.py
```diff
@@ -141,22 +178,76 @@ def _construct_proofs(
...
+            secrets
+        ), f"len(amounts)={len(amounts)} not equal to len(secrets)={len(secrets)}"
+        outputs: List[BlindedMessage] = []
+
+        rs_ = [None] * len(amounts) if not rs else rs
+        rs_return: List[PrivateKey] = []
+        for secret, amount, r in zip(secrets, amounts, rs_):
+            B_, r = b_dhke.step1_alice(secret, r or None)
+            rs_return.append(r)
+            output: BlindedMessage = BlindedMessage(
```
xphade, line 220: [nitpick] I guess the type hint is a bit redundant here.

cashu/wallet/wallet.py
```diff
@@ -141,22 +178,76 @@ def _construct_proofs(
...
+                amount=amount, B_=B_.serialize().hex()
+            )
+            outputs.append(output)
+        for o, r in zip(outputs, rs_return):
+            logger.trace(f"Constructing output: {o}, r: {r.serialize()}")
```
xphade, line 225: [nitpick] Couldn't you just remove this additional loop and move it to the one above?

cashu/wallet/wallet.py
```diff
@@ -561,29 +642,244 @@ def _meltrequest_include_fields(proofs):
...
         except Exception as e:
-            pass
+            logger.error(e)
+
+    async def _generate_secret(self, randombits=128) -> str:
+        """Returns base64 encoded deterministic random string.
+
+        NOTE: This method should probably retire after `deterministic_secrets`. We are
+        deriving secrets from a counter but don't store the respective blinding factor.
+        We won't be anle to restore any ecash generated with these secrets.
```
xphade, line 781: typo: able

cashu/wallet/cli/cli.py
```diff
@@ -444,6 +480,47 @@ async def burn(ctx: Context, token: str, all: bool, force: bool, delete: str):
...
 
 
+@cli.command("restore", help="Restore backups.")
+@click.option(
+    "--to",
+    "-t",
+    default=2,
+    help="Split minted tokens with a specific amount.",
+    type=int,
+)
```
xphade, line 485: Maybe it's just me being not too deep into the code but it isn't quite clear to me what this `to` option / variable is doing.  Also at the other places where it is used (I think router.py and wallet.py). 

Could we maybe improve the name or the docstrings a bit to make it better understandable what this is doing?

xphade (commented):
Hey @callebtc, very cool new feature, I really like it.

I had a high-level look through the code and it looks good to me, just some minor comments. I probably don't yet understand all implications of this change though. :slightly_smiling_face: 

One additional question, from your description:

>  Should I derive secret from r or generate them independently?

Looks like you decided to generate them separately. Was there any specific reason for this or just a random choice?

Also, this will probably become a NUT, right?

cashu/wallet/cli/cli.py
callebtc: Yes, deprected because `cli` is now `async` :)

cashu/wallet/cli/cli.py
callebtc: Yes, this line was deprecated as well. The comment relates to the new `cls` method `.with_db()` where we need to `skip_private_key` once in cases where the db wasn't migrated yet. Not ideal solution but still a lot better than before. I think this can be cleaned up later.

cashu/wallet/cli/cli.py
callebtc: Oh yeah, nice catch. The docstring is totally wrong. This flag specifies how many batches should be attempted to be restored after no new proof is found. I think users should never make use of this (the default value should just work) but I left it in for experimenting 
```suggestion
    default=2,
    help="Split minted tokens with a specific amount.",
    type=int,
)
@click.option(
    "--batch",
    "-b",
```until we figure out what the best parameters for this should be.

**Comment by callebtc** [comment]
> Hey @callebtc, very cool new feature, I really like it.
> 
> I had a high-level look through the code and it looks good to me, just some minor comments. I probably don't yet understand all implications of this change though. :slightly_smiling_face: 
> 
> One additional question, from your description:
> 
> >  Should I derive secret from r or generate them independently?
> 
> Looks like you decided to generate them separately. Was there any specific reason for this or just a random choice?
> 
> Also, this will probably become a NUT, right?


Hey there, yes at least for now `r` is independently derived, and not derived from `x`.

There are intricacies that arise if you derive one from the other. In order to keep those out, I've decided on this approach.

I will write a NUT for this for sure after it has proven to work well!

File: cashu/wallet/wallet.py [hunk 11]
```diff
@@ -539,29 +604,259 @@ def _meltrequest_include_fields(proofs):
...
+            # or use the one provided
+            mnemonic_str = from_mnemonic
+        elif ret_db is not None:
+            # if there is a seed in the database, use it
+            _, mnemonic_str = ret_db[0], ret_db[1]
+        else:
+            logger.debug("No mnemonic provided")
+            return
+
+        if not mnemo.check(mnemonic_str):
```
Mynima, line 728: Consider adding .lower() to the seed when it is used so that the re-entry of the seed isn't case sensitive. No need for it to be. Tweaked my local copy and it was working nicely without throwing the "Invalid mnemonic" for mixed casing seeds.

Also maybe a bit more complex but could be useful to allow seed restoration with just first 4 (or 3) characters as no need for the user to write them all out. Maybe this is a nice-to-have though, if it is a lot of work maybe not worth it.

Mynima:
Great work, I tried to break it and found it difficult to do so. Will try put some eyes on the code, though with the codebase being quite mature as a whole it is difficult to fully create a mental model without substantial time reviewing/dabbling.

Main test that found inconsistent behavior was around the initiation of restoring wallets while there were pending ecash sends at play. Sometimes able to receive previous sends even after wallet is restored (simulating alternative device). This leads to confusion and inability to accurately get pending information. It also can result in it being difficult to initiate new sends potentially pinning the funds (temporarily?).

Also performed test for 1000 cycles of invoicing (1sat per cycle) from test mint. Got to around 700ish and then started to receive errors from the mint about the invoice secrets already being used.

Test with around 100 iterations of invoices and some sends restored and balance reflected correctly.

Similar test with 700+ iterations showed balance of funds as expected. Removed DB and restored by balance showed 0 on restore. This item could be related to the test mint. Does it have some time limit/restrictions on the tokens minted? (it did take some time to cycle through the 700+ iterations).

Some restorations showed 0 balance but then other times after some time restoration showed funds. Again this could be related to the test mint if there are some odd interactions with mint time limits or if the funds get swept after some time.

callebtc:
Hello @Mynima thank you for that awesome detailed response. Could you share the exact steps and set of commands that lead to some of the inconsistencies you have described? 

I'm not sure if there is an easy fix for restoring when pending tokens are in the db. The tokens don't have a special state from the perspective of the mint so a restore should make them available in the new wallet as if they are normal non-pending tokens again. If that was the case for you, everything should be fine.

As I said, I would love to be able to reproduce the other behavior you described. If you can help me with those, I should be able to write tests that reveal these inconsistencies. Cheers!

Mynima:
> Hello @Mynima thank you for that awesome detailed response. Could you share the exact steps and set of commands that lead to some of the inconsistencies you have described?
> 
> I'm not sure if there is an easy fix for restoring when pending tokens are in the db. The tokens don't have a special state from the perspective of the mint so a restore should make them available in the new wallet as if they are normal non-pending tokens again. If that was the case for you, everything should be fine.
> 
> As I said, I would love to be able to reproduce the other behavior you described. If you can help me with those, I should be able to write tests that reveal these inconsistencies. Cheers!

No worries at all, it was definitely doing something funny where it appeared that funds (after restore) were pinned (perhaps temporarily) from the perspective of performing a 'SEND ecash' action.

Here is the sequence of events to re-test:
1. Create two new wallets (call wallet_a and wallet_b)
2. Fund wallet_a with an invoice of 200 sats
3. Create ecash send from wallet_a of 100sat (save token temporarily and do not commit to a receive)
4. Restore wallet_a into a new wallet_c
5. Check balance wallet_c (should be 200 sats)
6. Attempt to perform a send 100 sats from wallet_c (result should be error message)
7. Receive wallet_a send to wallet_b and check balances (wallet_b should have 100 sats)
8. Attempt to perform a send 100 sats from wallet_c (result should be error message, repeat for smaller amounts until successful)

You may want to repeat this test a few times as sometimes it looked like it could be OK and the funds were available in the restored wallet. The situation this simulates is if someone has tried to send funds and then restored or dual restored with a second device. 

By offering recovery we inherently open the door to people thinking about having the same seed on multiple devices, as it is how we're conditioned to think about bitcoin. However, the state of the ecash is governed by the private DBs and not a shared ledger, which means that restoring in multiple places does give you access to the initial pool of funds, however each devices/restore will have a separate database on which the state of the wallet will be tracked. This means we need to provide clear guidance that users should consider restoration as a means of moving devices (or retrieval of lost funds), not means of having multiple installations. It would be more advisable that for multiple devices users maintain separate wallets and send/receive between these wallets as needed.

For the other tests I wrote a simple loops in the shell. 
1. Firstly to 1000 sat invoices requested for a new wallet. That should be an easy test to do. 
2. Once this loop is finished (if no errors about the secret already being used were found) a few send/receives should be sent to another wallet, the balance then checked. 
3. Then forget/remove the DB
4. Restore from seed (making sure as many restore iterations are completed as necessary)
5. Check balance (this should match step 2)

callebtc:
> * Attempt to perform a send 100 sats from wallet_c (result should be error message)
> * Receive wallet_a send to wallet_b and check balances (wallet_b should have 100 sats)

Gotcha, I think here lies the crux: When you do a `cashu --wallet wallet_a send 100`, what happens is that 100 sats are burned from the wallet (now spent!) and replaced with a new set of 100 sats that are then stored in the wallet with the state "pending". These are "normal" funds that can be received by anyone, including yourself. They are not "locked" from the perspective of the mint.

To your first point: Now, when you restore `wallet_a` to `wallet_c` and `cashu send --wallet=wallet_c 100`, the expected behavior is that it _should not_ error because you are spending from the funds that are marked "pending" in `wallet_a`. I'm trying to say that `wallet_c` has no way of knowing that these 100 sats are "pending", neither does the mint.

To your second point: When you try to send the "pending" 100 sats from `wallet_a` to `wallet_c` something weird happens: you are trying to receive funds in `wallet_c` that you already have in your database. That's why your balance remains at 200 (200 your initial balance and 100 sats that are duplicate, being a total of 200 unique ecash). 

Hope it makes sense. Very interesting and weird use case you have there. It will cause weird problems when you use the same seed across two wallets and I can't think of a way to prevent that.

callebtc:
> By offering recovery we inherently open the door to people thinking about having the same seed on multiple devices, as it is how we're conditioned to think about bitcoin. However, the state of the ecash is governed by the private DBs and not a shared ledger, which means that restoring in multiple places does give you access to the initial pool of funds, however each devices/restore will have a separate database on which the state of the wallet will be tracked. This means we need to provide clear guidance that users should consider restoration as a means of moving devices (or retrieval of lost funds), not means of having multiple installations. It would be more advisable that for multiple devices users maintain separate wallets and send/receive between these wallets as needed.

Perfect summary 🙏

callebtc:
Thank you to all reviewers! LFG!

Issue #116, opened 2023-03-04
[Wallet] Todo: Improve `cashu pending` and `cashu burn` and `cashu receive`  functionality for pending tokens


callebtc:
Glossary: 
- _pending_: an outgoing token that has not been received 
- _burn_: delete an outgoing token that has been received from the db
- _redeem_: to receive a token by the receiving party

### Background

The command `cashu pending` shows all created (outgoing) tokens that have not been _recognized as redeemed_ (i.e. received by the other party) yet. When users send a token with `cashu send <amount>` the client gives each sending attempt a unique _send ID_ to be able to later sort them and display them in the `cashu pending` view. This view is useful because we can keep track of all tokens and _burn_ them (i.e. delete from from our database) after they have been received. Another reason why we need to display pending tokens is to be able to receive them back to yourself, if the receives is unavailable (or you want to cancel an incomplete transaction).

### Burning tokens 

To burn a specific token, the user can enter `cashu burn <token>`. A token will only be burnt if it has been received, so this command should always be safe to execute. To burn all pending tokens we can use the command `cashu burn -a` and to burn _all_ tokens in our database, we can use `cashu burn -f`. Again, all of these commands should be safe to use. I think this works quite ok.

### Canceling transactions

Canceling transactions means receiving a pending token yourself. Right now, the user needs to enter `cashu pending` then copy each token they want to cancel one by one and then enter `cashu receive <token>` to receive it themselves. I think an option to cancel all pending tokens would be useful, i.e., something like `cashu receive -a`, which would be equivalent to `cashu burn -a`.

### Improvements

- Ability to show only n pending tokens (with pagination/offset), like `cashu pending -n 10 --offset 50` would show pending tokens with index -50 to -60.
- Ability to receive all pending tokens `cashu receive -a`
- Ability to force-delete entries from the pending token list (by send ID) in cases where the mint is not available anymore

PR #135, opened 2023-03-07
Add ability to show n pending tokens starting from offset

sihamon:
See #116.
`cashu pending -n 10 --offset 50` will show 10 pending tokens starting from index 50.


callebtc:
Great work! LGTM

PR #142, opened 2023-03-13
Add ability to force-delete pending token by send ID

sihamon:
This might be useful in cases the mint becomes unavailable. See #116.
To remove a pending token from the list use `cashu burn --delete send-id`. 
The token is forcefully deleted without further checking with the mint (i.e. if token is already spent) also in cases where the mint would still be available.


callebtc:
LGTM

Tested, works great!

PR #147, opened 2023-03-16
Enable postgres

trbouma:
postgres support. I require this for my deployment. I uncommented the code and it works fine with my postgres server.


trbouma:
another pull request with Docker changes

PR #148, opened 2023-03-19
Postgres with Dockerfile changes

trbouma:
Here is a new pull request with the Postgres addition with a change to the Docker file so that it builds. I deleted the other pull request, please use this one.


callebtc:
LGTM

PR #150, opened 2023-03-20
Description Hash required for Lightning Wallet Integration

trbouma:
This is needed for a tighter Lightning Wallet Integration with Cashu Mint. When the callback is returned to the wallet, the metadata_str that is provided is used to generated a description_hash which is used to request the lightning invoice that is generated by the mint. When the invoice is returned, the Lightning Wallet checks to see if the description_hash is the invoice is the same before proceeding, otherwise aborts. This was required to get past step 7 of lud-06 which had me stumped for quite a while, and this PR fixes it. 

In the cashu mint, the description_hash needs to be added in the parameters for the functions that pass the data and generate the invoice, namely ledger.py, router.py, and wallet.py. I also discovered there is an encoding bug in lnbits.py, that likely never got caught, because until now, description_hash was always None.

Also a few minor changes to the Dockerfile and poetry files to add a dependency. 

I have the PR built as docker image and running in my system. Happy to answer any questions!


trbouma:
@callebtc - no problem closing this pull request, however I need this capability for my use case and will continue to keep my branch up to date with latest commits on main. I need to better understand how this supports or conflicts with LNURL, and what the implications might be and make a stronger case to have this merged into main.

PR #152, opened 2023-03-26
Fix: TypeError in burn token

sihamon:
When burning specific token via cashu burn "token", there was a TypeError in line 459 of cli.py:
`proofs = [Proof(**p) for p in json.loads(base64.urlsafe_b64decode(token))]`

`
TypeError: cashu.core.base.Proof() argument after ** must be a mapping, not str
`


callebtc:
Nice. I'll merge this and make a PR with it using `token = wallet._deserialize_token_V3(token)` instead. Thank you.

Issue #39, opened 2022-10-17
Bounty: Create a REST API for the Cashu Wallet Client


Altbierjupp:
This Issue is to create a REST API for the Cashu Wallet Client for easier access of the functions included inside the wallet. The idea is to extend the wallet with an API to make it much easier to integrate in other applications and developments.

More details can be added inside this issue!

This is a Bounty offered by the [Nodesignal Podcast](https://twitter.com/nodesignal) and the developer who provide this functionality can get 250,000 Sats as Bounty. More Information about the Bounty program can be found here:
[https://nodesignal.space/spenden-bitcoin-bounties/](https://nodesignal.space/spenden-bitcoin-bounties/)

callebtc:
## Bounty goals

The goal of this bounty is to extend the Cashu python [wallet](https://github.com/callebtc/cashu/tree/main/cashu/wallet) such that it can run in a `daemon` mode in the background and offer at least the same functionality as the CLI to a client REST API and thus to other applications running in the same network. We suspect that a successful execution of this idea will make it easier to build GUI wallet frontends around the Cashu wallet daemon since all token management and mint communication can be entirely left to the daemon and the wallet frontend can focus on UI and UX. 

In the best case and as a bonus, the submissions for this bounty could be bundled with a very simple example of a  Javascript UI packaged together with the Python wallet app such that both can be shipped to a user in the future.

The API integration should be executed via `uvicorn` and `FastApi` since those two frameworks are already used in Cashu.

Note: This description is a work-in-progress and could change a bit in the future.

trbouma:
This is what I have so far. I have added the capability of managed wallets. app.anonsats.io

callebtc:
Congratulations for the bounty @sihamon!

PR #154, opened 2023-03-30
Add REST API for Cashu wallet

sihamon:
See #39.
- add an API endpoint for every CLI command
- add necessary helper functions for API (including TokenV3), these could be unified with helper functions for CLI in the future
- add simple tests for API


callebtc:
Amazing work! You're reimplemented almost the entire CLI which is impressive. This approach will make it a lot harder to maintain CLI and API at the same time. Do you think there is any way we could refactor things that both CLI and API use the same set of functions? See `cli_helpser.py` for example. Something like that could maybe be used by API and CLI.

I think the overall goal is to have as few lines of code as possible. Should we refactor the business logic out of the CLI and reuse it in the API?

callebtc:
Amazing job. Tests failing is very likely my fault due to the change of the default derivation path. I will fix that and review your PR asap!

cashu/wallet/helpers.py
```diff
@@ -0,0 +1,453 @@
...
+    print(mint_url)
+    print(wallet.name)
+    # load this mint_url into a wallet
+    mint_wallet = Wallet(mint_url, os.path.join(settings.cashu_dir, wallet.name))
+    mint_keysets: WalletKeyset = await get_keyset(mint_url=mint_url, db=mint_wallet.db)  # type: ignore
+    print(mint_keysets)
+    print(mint_wallet)
+    # load the keys
+    assert mint_keysets.id
+    await mint_wallet.load_mint(keyset_id=mint_keysets.id)
```
callebtc, line 257: Needs to be rebased to #162 because of a bug (and some forgotten comments seem to be left).

cashu/wallet/cli/cli.py
```diff
@@ -87,7 +69,9 @@ def cli(ctx: Context, host: str, walletname: str):
     ctx.ensure_object(dict)
     ctx.obj["HOST"] = host
     ctx.obj["WALLET_NAME"] = walletname
-    wallet = Wallet(ctx.obj["HOST"], os.path.join(settings.cashu_dir, walletname))
+    wallet = Wallet(
+        ctx.obj["HOST"], os.path.join(settings.cashu_dir, walletname), name=walletname
```
callebtc, line 73: Does adding `name=walletname` do anything?

cashu/wallet/cli/cli_helpers.py
```diff
@@ -1,252 +0,0 @@
...
-        return
-    proofs_keysets = set([p.id for p in token.proofs])
-
-    logger.debug(f"Verifying mints")
-    trust_token_mints = True
-    for mint in token.mints:
-        for keyset in set([id for id in mint.ids if id in proofs_keysets]):
-            # init a temporary wallet object
-            keyset_wallet = Wallet(
-                mint.url, os.path.join(settings.cashu_dir, ctx.obj["WALLET_NAME"])
```
callebtc, line 70: Now I see why you set the wallet name!

cashu/wallet/helpers.py
```diff
@@ -0,0 +1,453 @@
...
+    Helper function that prints the balances for each mint URL that we have tokens from.
+    """
+    # get balances per mint
+    mint_balances = await wallet.balance_per_minturl()
+
+    # if we have a balance on a non-default mint, we show its URL
+    keysets = [k for k, v in wallet.balance_per_keyset().items()]
+    for k in keysets:
+        ks = await get_keyset(id=str(k), db=wallet.db)
+        if ks and ks.mint_url != wallet.url:
```
callebtc, line 195: I see the same use of `wallet.url` instead of `ctx.obj["HOST"]` here. Nice.

cashu/wallet/helpers.py
```diff
@@ -0,0 +1,453 @@
...
+            _, _ = await mint_wallet.redeem(proofs, script, signature)
+            if not is_api:
+                print(f"Received {sum_proofs(proofs)} sats")
+
+    # reload main wallet so the balance updates
+    await wallet.load_proofs()
+    if is_api:
+        return wallet.available_balance
+    else:
+        wallet.status()
```
callebtc, line 403: Function should not return two different types.

cashu/wallet/helpers.py
```diff
@@ -0,0 +1,453 @@
...
+                assert mint_keysets.mint_url, Exception("we don't know this mint's URL")
+            # now we have the URL
+            mint_wallet = Wallet(
+                mint_keysets.mint_url,
+                os.path.join(settings.cashu_dir, wallet.name),
+            )
+            await mint_wallet.load_mint(keyset_in_token)
+            _, _ = await mint_wallet.redeem(proofs, script, signature)
+            if not is_api:
+                print(f"Received {sum_proofs(proofs)} sats")
```
callebtc, line 396: I think the print can remain also if `is_api`.

cashu/wallet/helpers.py
```diff
@@ -0,0 +1,453 @@
...
+            mint_keysets = await get_keyset(id=keyset_in_token, db=wallet.db)
+            if is_api:
+                assert mint_keysets, HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="we don't know this keyset",
+                )
+                assert mint_keysets.mint_url, HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="we don't know this mint's URL",
+                )
```
callebtc, line 384: Lot's of repeated code due to error handling. Why wouldn't it be better if the API at `api/app.py` simply caught any exceptions thrown in `helpers.py` and then return a `HTTPException`.

There should be no `HTTPException` in this file ideally. What do you think?

cashu/wallet/helpers.py
```diff
@@ -0,0 +1,453 @@
...
+            print("")
+        token = await wallet.serialize_proofs(
+            send_proofs,
+            legacy=True,
+        )
+        if not is_api:
+            print(token)
+
+    if is_api:
+        return wallet.available_balance, token
```
callebtc, line 451: All prints can remain also for `is_api`.

Ideally, there should be no `is_api` in here since it hopefully doesn't change the business logic anyway.

cashu/wallet/helpers.py
callebtc: What would be better is to leave functions in `cli_helpers` or `helpers.py` as unmodified as possible (except maybe for the contexts) but handle all changes to business logic (in this case you want `wallet.status()`) outside of these files, somewhere in `api/`.

callebtc:
I gave the PR another round of eyes. A lot of code has gone from `cli_helpers.py` to `helpers.py`. Unfortunately, it's not easy to compare the changes. When you look at it in the github diff, It's just a huge block of code that moved. It's hard for me to see what the new parts are.

Here are a few ideas of refactoring things a bit better:
- I see you replaced all `ctx.` access to `wallet.`. This happened for wallet name and url. I think this is a good approach. Did you do it for anything else? Hard to find all changes. 
- Some functions in `helpers.py` return multiple types, depending on `is_api`. This is not good. Functions should ideally return always the same type if possible to keep code simpler. Functions in `helpers` should remain as unchanged as possible and should be called in `api/`. Everything that happens because of `is_api` should remain in `api/` as much as possible. (An exception to this is the changes to `ctx` above).
- The same idea for exception handling: `helpers.py` should not return `HTTPException`s but only normal ones. These need to be caught in the parent in `api/` and  returned as HTTP Exceptions.

Thank you again for the fantastic work you've done. I'm really impressed.

callebtc:
Does rebasing to this PR #163 make your life very complicated?

cashu/wallet/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+        includes_mint_info: bool = any([t.mint for t in tokenObj.token])
+
+        # if there is a `mints` field in the token
+        # we check whether the token has mints that we don't know yet
+        # and ask the user if they want to trust the new mints
+        if includes_mint_info:
+            # we ask the user to confirm any new mints the tokens may include
+            # await verify_mints(ctx, tokenObj)
+            # redeem tokens with new wallet instances
+            await redeem_TokenV3_multimint(
```
callebtc, line 336: Oooof. Almost missed this. The entire code block has wrong indentation.

callebtc:
I think there is an opportunity here to refactor the mint selection out of the helpers into `cli.py` and fix #170 at the same time.

PR #162, opened 2023-04-16
fix: load mint with most recent keyset

callebtc:
Wallet used to load the mint with the keyset_id it encountered first in the database instead of loading the most recent one which is determined by asking the mint directly because we use `mint_wallet.load_mint(keyset_id=None)` now.

### DIFF
--- a/cashu/wallet/cli/cli_helpers.py
+++ b/cashu/wallet/cli/cli_helpers.py
@@ -185,11 +185,11 @@ async def get_mint_wallet(ctx: Context):
     Helper function that asks the user for an input to select which mint they want to load.
     Useful for selecting the mint that the user wants to send tokens from.
     """
+    # we load a dummy wallet so we can check the balance per mint
     wallet: Wallet = ctx.obj["WALLET"]
-    await wallet.load_mint()
-
     mint_balances = await wallet.balance_per_minturl()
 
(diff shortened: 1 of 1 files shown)


cashu/wallet/cli/cli_helpers.py
```diff
@@ -185,11 +185,11 @@ async def get_mint_wallet(ctx: Context):
     Helper function that asks the user for an input to select which mint they want to load.
     Useful for selecting the mint that the user wants to send tokens from.
     """
+    # we load a dummy wallet so we can check the balance per mint
     wallet: Wallet = ctx.obj["WALLET"]
-    await wallet.load_mint()
```
callebtc, line 189: This was a redundant step.

cashu/wallet/cli/cli_helpers.py
```diff
@@ -215,11 +215,7 @@ async def get_mint_wallet(ctx: Context):
     mint_wallet = Wallet(
         mint_url, os.path.join(settings.cashu_dir, ctx.obj["WALLET_NAME"])
     )
-    mint_keysets: WalletKeyset = await get_keyset(mint_url=mint_url, db=mint_wallet.db)  # type: ignore
```
callebtc, line 218: This uses a `fetchone` lookup which means that the wallet will load the first keyset_id it will get from the db.

PR #172, opened 2023-04-21
fix double "or" typo in burn parameter validation error message

theStack:
While making myself familiar with some of the available cashu commands, I noticed that `burn` command parameter validation error message contains "oror" instead of "or". (Found my way here through WBD episode #647, the project looks very promising!).


callebtc:
Good catch, thank you! Hope you'll look more into it!

PR #174, opened 2023-04-23
fix: test_step2

ngutech21:
The test_step2() doesn't call step2_bob() of the b_dkhe module and therefore is only testing step1_alice()


callebtc:
Nice  catch, thank you! LGTM

PR #182, opened 2023-04-28
Fix: invalid public key

moonsettler:
```
        # DLEQ verify
        # the below lines break the format of C and result in 'invalid public key'
        # error when the mint tries to verify it's blind signature.
        # unclear what the purpose of this exercise was, it's not fit for negative
        # test in the DLEQ proofs, so i commented them out - moonsettler
        #
        # print(f"before:{scnd_proofs[0].C}")
        # scnd_proofs[0].C = "1" + scnd_proofs[0].C[1:]
        # print(f"after: {scnd_proofs[0].C}")
```

### DIFF
--- a/cashu/wallet/wallet.py
+++ b/cashu/wallet/wallet.py
@@ -572,9 +572,14 @@ class Wallet(LedgerAPI):
             raise Exception("received no splits.")
 
         # DLEQ verify
-        print(f"before:{scnd_proofs[0].C}")
-        scnd_proofs[0].C = "1" + scnd_proofs[0].C[1:]
-        print(f"after: {scnd_proofs[0].C}")
+        # the below lines break the format of C and result in 'invalid public key'
+        # error when the mint tries to verify it's blind signature.
+        # unclear what the purpose of this exercise was, it's not fit for negative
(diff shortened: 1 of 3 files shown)


cashu/wallet/wallet.py
```diff
@@ -572,9 +572,14 @@ async def split(
...
-        scnd_proofs[0].C = "1" + scnd_proofs[0].C[1:]
-        print(f"after: {scnd_proofs[0].C}")
+        # the below lines break the format of C and result in 'invalid public key'
+        # error when the mint tries to verify it's blind signature.
+        # unclear what the purpose of this exercise was, it's not fit for negative
+        # test in the DLEQ proofs, so i commented them out - moonsettler
+        #
+        # print(f"before:{scnd_proofs[0].C}")
+        # scnd_proofs[0].C = "1" + scnd_proofs[0].C[1:]
+        # print(f"after: {scnd_proofs[0].C}")
```
callebtc, line 582:
```suggestion
```

callebtc: let's get rid of it, seems like a debug thing I left in there

PR #194, opened 2023-05-08
Mint working with Postgres

trbouma:
This is the branch I am using to run the mint with postgres. I don't expect you to merge, but you can inspect the changes I have been making.

1. Views were broken. I fixed by moving the 'AS' sub-clause to where postgres likes it. Also, postgress doesn't like the select clauses so I hacked to get past the errors.
2. Timestamps - postgres is looking for a different format, so I complied with using the datetime module.

This is all working in my deployment with postgress, but I haven't yet tested back onto SQLite. Also, an unexpected benefit to this is that now I can use postgres database cluster to backend user wallets instead of sqlite files. sqlite is only designed to work with local filestores, so moving to postgres is a big scalability bonus.

Also, I still have the description_hash parameters - I need this for lightning wallet integration.


cashu/wallet/migrations.py
```diff
@@ -43,21 +43,21 @@ async def m001_initial(db: Database):
...
         SELECT COALESCE(SUM(s), 0) AS used FROM (
-            SELECT SUM(amount) AS s
+            SELECT SUM(amount) 
             FROM proofs_used
-            WHERE amount > 0
-        );
+            WHERE amount > 0 
+        ) AS s;
     """
     )
```
callebtc, line 63: I believe this is the fix and the rest is accidentally committed code?

trbouma: @callebtc  Yes, that is the fix. You can ignore the other changes for now, because those are mods to my desc_hash branch which should not be merge. 

I also hacked a fix on another view that is trying to a calculation in the SELECT field. Postgres didn't like it so I removed the second operand so the SQL statement goes through. It was for calculating the mint balance_issued-balance_redeemed. It didn't seem to be used there, so I made the change to make my stuff work.

trbouma:
I opened another with just the key changes

PR #208, opened 2023-05-13
Assert `mint_private_key` is available at startup

### LINKED ISSUE #179: [Mint] Error: AttributeError: 'NoneType' object has no attribute 'encode'
`AttributeError: 'NoneType' object has no attribute 'encode'`

when .env file is not present and running

```bash
python -m cashu.mint
```

xphade:
If the mint's private key is not available, the ledger cannot be created because no pubkey can be derived. We now explicitly assert this with a descriptive error message.

Additionally fixed some typing errors.

Closes #179

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,9 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Tor PID
+tor.pid
+
+# Default data directory
+/data
--- a/cashu/mint/startup.py
+++ b/cashu/mint/startup.py
@@ -21,6 +21,10 @@ for key, value in settings.dict().items():
 wallets_module = importlib.import_module("cashu.lightning")
 lightning_backend = getattr(wallets_module, settings.mint_lightning_backend)()
 
+assert (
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -127,3 +127,9 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Tor PID
+tor.pid
+
+# Default data directory
+/data
```
xphade, line 135: Let me know if these were not in the ignore file on purpose.

cashu/mint/startup.py
```diff
@@ -21,6 +21,10 @@
 wallets_module = importlib.import_module("cashu.lightning")
 lightning_backend = getattr(wallets_module, settings.mint_lightning_backend)()
 
+assert (
+    settings.mint_private_key is not None
+), "Cannot create ledger: No private key set for mint. Check your environment settings."
```
callebtc, line 26:
```suggestion
), "no private key is set."
```

callebtc:
LGTM

PR #212, opened 2023-05-13
Add multi-mint swap to CLI and API

### LINKED ISSUE #171: [Wallet] Feature request: Multi-mint swaps
Command to let the user send funds from one mint to another. User could be asked to select an outgoing mint, then an incoming mint, and then specify the amount and confirm the transaction.

sihamon:
Add command `swap` to CLI and endpoint `swap` to API to enable swaps between different mints.
Closes #171 .


cashu/wallet/api/router.py
```diff
@@ -123,6 +123,50 @@ async def invoice(
...
+    incoming_mint: str = Query(default=..., description="URL of incoming mint"),
+):
+    assert settings.lightning, HTTPException(
+        status_code=status.HTTP_400_BAD_REQUEST,
+        detail="lightning not enabled",
+    )
+    assert outgoing_mint not in incoming_mint, HTTPException(
+        status_code=status.HTTP_400_BAD_REQUEST,
+        detail="mints for swap have to be different",
+    )
```
AngusP, line 139: Using `assert` will crash the server completely, no? Instead these should be `raise`:

```suggestion
    if not settings.lightning:
        raise HTTPException(
            status_code=status.HTTP_400_BAD_REQUEST,
            detail="lightning not enabled",
        )
    if not outgoing_mint not in incoming_mint:
        raise HTTPException(
            status_code=status.HTTP_400_BAD_REQUEST,
            detail="mints for swap have to be different",
        )
```

Would also suggest `"This mint does not support Lightning"` instead of `"lightning not enabled"` as it's more descriptive for a user?

(Also could use **501 Not Implemented** or **403 Forbidden** for the first check ?)

AngusP: Unless `AssertionError` gets caught somewhere but almost always that's a no-no

callebtc: I actually thought you can't raise custom exceptions with an `assert`. However, it will not crash the server, just the thread on which the endpoint lives.

cashu/wallet/api/router.py
```diff
@@ -139,6 +140,58 @@ async def invoice(
...
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="mints for swap have to be different",
+        )
+    global wallet
+    # request invoice from incoming mint
+    wallet = await load_mint(wallet, incoming_mint)
+    invoice = await wallet.request_mint(amount)
+    # pay invoice from outgoing mint
+    wallet = await load_mint(wallet, outgoing_mint)
+    await wallet.load_proofs()
```
AngusP, line 170: Question: If `wallet` is a **global**, and the mint is changed here, during this time any other requests to the wallet API could accidentally get the outgoing mint as the loaded mint?

Additionally, another request could itself change the mint? i.e. if I made two rapid requests to `/swap`, there could be four different mints involved and the swaps could mess each-other up?

cashu/wallet/cli/cli.py
```diff
@@ -171,6 +171,57 @@ async def invoice(ctx: Context, amount: int, hash: str):
...
+@cli.command("swap", help="Swap funds between mints.")
+@click.pass_context
+@coro
+async def swap(ctx: Context):
+    if not settings.lightning:
+        raise Exception("this mint does not support lightning.")
+    print("Mint to swap from:")
+    wallet_outgoing_mint = await get_mint_wallet(ctx)
+    amount = int(input("Enter amount to swap in sats:"))
+    incoming_mint_url = input("Enter URL of mint to swap to:")
```
callebtc, line 183: why not load the mint like here?
```python
wallet_outgoing_mint = await get_mint_wallet(ctx)
```

cashu/wallet/api/router.py
```diff
@@ -127,6 +128,46 @@ async def invoice(
...
+    summary="Swap funds between mints.",
+    response_model=SwapResponse,
+)
+async def swap(
+    amount: int = Query(default=..., description="Amount to swap between mints"),
+    outgoing_mint: str = Query(default=..., description="URL of outgoing mint"),
+    incoming_mint: str = Query(default=..., description="URL of incoming mint"),
+):
+    if not settings.lightning:
+        raise Exception("this mint does not support lightning.")
```
AngusP, line 143: This setting applies to the local wallet/mint, but for a swap it only matters whether the incoming mint supports Lightning? Which I think you can only check if it does by requesting a mint invoice and seeing if it responds with one or an error, as this isn't specified in `/info`.

Could then avoid re-loading the incoming mint on L159 
```python
    # mint token in incoming mint
    wallet = await load_mint(wallet, incoming_mint)
```

cashu/wallet/api/router.py
```diff
@@ -127,6 +128,46 @@ async def invoice(
...
+    if outgoing_mint in incoming_mint:
```
AngusP, line 144: Why `in`? It'd be a bit odd but two distinct mints could be running at different URLs where one is a substring of the other, e.g. `https://example.com/cashu` and `https://example.com/cashu/staging`. Also weird but the same mint could also run at two different URLs (perhaps Tor/clearnet would be the most likely way to see this)

I *think* the only way to check if the mints are different is if they give a different `pubkey` in their `/info` response

cashu/wallet/cli/cli.py
```diff
@@ -171,6 +171,57 @@ async def invoice(ctx: Context, amount: int, hash: str):
...
+    if wallet.available_balance < total_amount:
+        raise Exception("balance too low")
+    _, send_proofs = await wallet.split_to_send(wallet.proofs, total_amount)
+    await wallet.pay_lightning(send_proofs, invoice.pr, fee_reserve_sat)
+    # mint token in incoming mint
+    wallet = Wallet(
+        incoming_mint_url,
+        os.path.join(settings.cashu_dir, ctx.obj["WALLET_NAME"]),
+        ctx.obj["WALLET_NAME"],
+    )
```
AngusP, line 218: Similar here, could just create two wallets and keep them rather than shadowing `wallet`

```python
incoming_wallet = Wallet(
    incoming_mint_url,
    os.path.join(settings.cashu_dir, ctx.obj["WALLET_NAME"]),
    ctx.obj["WALLET_NAME"],
)
outgoing_wallet = Wallet(
    wallet_outgoing_mint.url,
    os.path.join(settings.cashu_dir, ctx.obj["WALLET_NAME"]),
    ctx.obj["WALLET_NAME"],
)
```

cashu/wallet/api/router.py
```diff
@@ -127,6 +128,46 @@ async def invoice(
...
+        raise Exception("mints for swap have to be different")
+    # request invoice from incoming mint
+    wallet = create_wallet(url=incoming_mint)
+    invoice = await wallet.request_mint(amount)
+    # pay invoice from outgoing mint
+    wallet = await load_mint(wallet, outgoing_mint)
```
AngusP, line 150: bit of a nitpick, but `incoming_wallet` and `outgoing_wallet` would be a lot clearer to follow than shadowing `wallet` for both, as two calls to `create_wallet` rather than using `load_mint`

```suggestion
    # request invoice from incoming mint
    incoming_wallet = create_wallet(url=incoming_mint)
    outgoing_wallet = create_wallet(url=outgoing_mint)
    invoice = await incoming_wallet.request_mint(amount)
    # pay invoice from outgoing mint
    await outgoing_wallet.load_proofs()
    total_amount, fee_reserve_sat = await outgoing_wallet.get_pay_amount_with_fees(invoice.pr)
```

AngusP (commented):
Some comments on initial checks then handling of `wallet`s, but the actual swap logic LGTM

AngusP (approved):

callebtc (approved):
LGTM fixed up the rest with latest changes from main

PR #220, opened 2023-05-15
Catch all exceptions in wallet API

### LINKED ISSUE #201: [Wallet API] Catch all `Exception` in FastApi
All `HttpExceptions()` could be converted to normal `Exception`s and FastApi can catch them. See here: https://stackoverflow.com/a/62407111

sihamon:
Move HTTPException from router to error handler and catch all exceptions in error handler.
Closes #201 .


cashu/wallet/api/error_handler.py
```diff
@@ -0,0 +1,17 @@
...
+    def get_route_handler(self):
+        original_route_handler = super().get_route_handler()
+
+        async def custom_route_handler(request: Request):
+            try:
+                return await original_route_handler(request)
+            except Exception as exc:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc)
+                )
```
AngusP, line 15: Would suggest

```suggestion
            except (AssertionError, KeyboardInterrupt):
                # These should never really be caught, 1st is "something real bad" and 2nd is a PITA if something gets stuck
                raise
            except Exception as exc:
                if isinstance(exc, HTTPException):
                    # Allows optionally setting a custom `status_code` if desired, but otherwise defaulting to below for convenience
                    raise
                raise HTTPException(
                    status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc)
                )
```

cashu/wallet/api/router.py
```diff
@@ -18,8 +18,10 @@
 from ...wallet.nostr import receive_nostr, send_nostr
 from ...wallet.wallet import Wallet as Wallet
 from .api_helpers import verify_mints
+from .error_handler import ErrorHandler
 
 router: APIRouter = APIRouter()
+router.route_class = ErrorHandler
```
AngusP, line 24: Hrm, did you consider a [Middleware](https://fastapi.tiangolo.com/tutorial/middleware/)? I suppose switching out the Route class would work too but a middleware is 'more normal'

cashu/wallet/api/router.py
```diff
@@ -32,7 +34,7 @@ async def load_mint(wallet: Wallet, mint: Optional[str] = None):
     try:
         await wallet.load_mint()
     except Exception as e:
-        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
+        raise Exception(str(e))
```
AngusP, line 37: Jokingly suggest using a custom Exception class `Cachuww` :sneezing_face:

cashu/wallet/api/error_handler.py
AngusP: `str(exc)` probably OK for now, but should be a little cautious about echoing the contents of whatever exception gets raised back to some random person on the internet, it can leak info about the system and make it easier to find vulnerabilities (minor example #228 where the API responses reveal that this mint was running Postgres and CLN specifically)

cashu/wallet/api/router.py
callebtc: `Cash'uwu`

cashu/wallet/api/router.py
```diff
@@ -32,7 +32,7 @@ async def load_mint(wallet: Wallet, mint: Optional[str] = None):
...
```
callebtc, line 35: This block can be replaced by simply `wallet.load_mint()` without `try: ... except: ...` – achieves the same thing.

callebtc (approved):
LGTM (pls check comment)

cashu/wallet/api/error_handler.py
AngusP: Would still recommend passing-up a `HTTPException` as-is if one is encountered, but could easilly be added later if someone needs this

```python
if isinstance(exc, HTTPException):
    # Allows optionally setting a custom `status_code` if desired, but otherwise defaulting to below for convenience
    raise
```

---

On the `AssertionError`, seems `assert` has been used in a few places so can pass on that suggestion (though IMHO even though `assert` is shorter, it is best reserved as a thing to actually cause a crash, like if somehow `1 != 1`)

PR #223, opened 2023-05-16
Implement dynamic amount of tokens for change

### LINKED ISSUE #183: [Wallet/Mint] Implement update to NUT-08: Dynamic amount of tokens for change
Instead of using 4 blank outputs, we should use the new dynamic amount. 

Todo:
- [ ] Mint
- [ ] Wallet
- [ ] Write test for wallet that checks expected returned fees?

xphade:
With the recent update to NUT-08, we can ensure that the amount of blank outputs is always enough to cover any overpaid lightning fees. This change implements this functionality for both the wallet and the mint. The mint updateis backwards-compatible with respect to old wallets.

Closes #183

### DIFF
--- a/cashu/core/helpers.py
+++ b/cashu/core/helpers.py
@@ -1,4 +1,5 @@
 import asyncio
+import math
 from functools import partial, wraps
 from typing import List
 
@@ -42,3 +43,13 @@ def fee_reserve(amount_msat: int, internal=False) -> int:
         int(settings.lightning_reserve_fee_min),
         int(amount_msat * settings.lightning_fee_percent / 100.0),
     )
+
+
+def calculate_number_of_blank_outputs(fee_reserve_sat: int):
+    """Calculates the number of blank outputs used for returning overpaid fees.
+
+    The formula ensures that any overpaid fees can be represented by the blank outputs,
+    see NUT-08 for details.
+    """
+    assert fee_reserve_sat > 0, "Fee reserve has to be positive."
+    return max(math.ceil(math.log2(fee_reserve_sat)), 1)
--- a/cashu/mint/ledger.py
+++ b/cashu/mint/ledger.py
@@ -515,7 +515,7 @@ class Ledger:
         total_provided: int,
         invoice_amount: int,
         ln_fee_msat: int,
-        outputs: List[BlindedMessage],
+        outputs: Optional[List[BlindedMessage]],
         keyset: Optional[MintKeyset] = None,
     ):
         """Generates a set of new promises (blinded signatures) from a set of blank outputs
@@ -523,15 +523,15 @@ class Ledger:
         fee reserve provided by the wallet and the actual Lightning fee paid by the mint.
 
         If there is a positive difference, produces maximum `n_return_outputs` new outputs
-        with values close or equal to the fee difference. We can't be sure that we hit the
-        fee perfectly because we can only work with a limited set of blanket outputs and
-        their values are limited to 2^n.
+        with values close or equal to the fee difference. If the given number of `outputs` matches
+        the equation defined in NUT-08, we can be sure to return the overpaid fee perfectly.
+        Otherwise, a smaller amount will be returned.
 
         Args:
             total_provided (int): Amount of the proofs provided by the wallet.
             invoice_amount (int): Amount of the invoice to be paid.
             ln_fee_msat (int): Actually paid Lightning network fees.
-            outputs (List[BlindedMessage]): Outputs to sign for returning the overpaid fees.
+            outputs (Optional[List[BlindedMessage]]): Outputs to sign for returning the overpaid fees.
 
         Raises:
             Exception: Output validation failed.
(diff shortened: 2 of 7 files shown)


cashu/core/helpers.py
```diff
@@ -42,3 +43,13 @@ def fee_reserve(amount_msat: int, internal=False) -> int:
...
+
+
+def calculate_number_of_blank_outputs(fee_reserve_sat: int):
+    """Calculates the number of blank outputs used for returning overpaid fees.
+
+    The formula ensures that any overpaid fees can be represented by the blank outputs,
+    see NUT-08 for details.
+    """
+    assert fee_reserve_sat > 0, "Fee reserve has to be positive."
+    return max(math.ceil(math.log2(fee_reserve_sat)), 1)
```
xphade, line 55: Not sure if I really should put it in here, or rather into the wallet directly.

tests/test_mint.py
```diff
@@ -110,3 +112,69 @@ async def test_generate_promises(ledger: Ledger):
...
+    invoice_amount = 100_000
+    fee_reserve = 2_000
+    total_provided = invoice_amount + fee_reserve
+    actual_fee_msat = 100_000
+
+    expected_returned_promises = 7  # Amounts = [4, 8, 32, 64, 256, 512, 1024]
+    expected_returned_fees = 1900
+
+    n_blank_outputs = calculate_number_of_blank_outputs(fee_reserve)
+    blinded_msgs = [step1_alice(str(n)) for n in range(n_blank_outputs)]
```
xphade, line 130: Should we rather use pre-computed blinded messages here? We don't use them for anything in this test, but maybe better to have it as deterministic as possible.

xphade:
I've added tests for the actual equation and the mint implementation. Let me know if we also should do something for the `Wallet.pay_lightning`. Wasn't sure how hard it is to test that, and the equation is pretty much the only thing that changed.

callebtc:
Excellent PR! Will review

cashu/mint/ledger.py
```diff
@@ -541,22 +541,24 @@ async def _generate_change_promises(
...
         logger.debug(
-            f"Lightning fee was: {ln_fee_sat}. User paid: {user_paid_fee_sat}. Returning difference."
+            f"Lightning fee was: {ln_fee_sat}. User paid: {user_paid_fee_sat}. "
+            f"Returning difference: {overpaid_fee_sat}."
         )
-        if user_paid_fee_sat - ln_fee_sat > 0 and outputs is not None:
-            # we will only accept at maximum n_return_outputs outputs
-            assert len(outputs) <= n_return_outputs, Exception(
-                "too many change outputs provided"
-            )
```
callebtc, line 555: Although not strictly necessary, for DOS safety, we could add a check here using `calculate_number_of_blank_outputs` to check whether the user has generated too many blank outputs.

tests/test_mint.py
callebtc: You can use a predetermined `blinding_factor` when calling `step1_alice`, then the outcome is determinstic!

tests/test_mint.py
```diff
@@ -110,3 +112,69 @@ async def test_generate_promises(ledger: Ledger):
...
+    )
+
+    assert len(promises) == expected_returned_promises
+    assert sum([promise.amount for promise in promises]) == expected_returned_fees
+
+
+@pytest.mark.asyncio
+async def test_generate_change_promises_legacy_wallet(ledger: Ledger):
+    # Check if mint handles a legacy wallet implementation (always sends 4 blank
+    # outputs) as well.
```
callebtc, line 146: Excellent! Very thoughtful!

callebtc (approved):
Beautiful PR! Pls see if my comments are of any benefit!

AngusP (approved):
LGTM, caveat that I'm still familiarizing with the code :wink:

tests/test_mint.py
xphade: > You can use a predetermined blinding_factor when calling step1_alice, then the outcome is determinstic!

Yeah, that was my proposal. I guess that's a yes from your side, so I'll add predefined blinding factors. :+1: 

> Actually, now I'm thinking this might break legacy wallets which always send 4 outputs (or we add an exception to that but that would be extra ugly).

Sorry, I'm  a bit confused here, can you expand on this? How does it  break legacy wallets? As you can see below, I also added a test for it.

cashu/mint/ledger.py
xphade: Sure, can do. What do you want to happen in that case? Right now, if a wallet sends too many blank outputs, I'm just ignoring the one that are not necessary, see line 558.

cashu/core/helpers.py
```diff
@@ -42,3 +43,13 @@ def fee_reserve(amount_msat: int, internal=False) -> int:
         int(settings.lightning_reserve_fee_min),
         int(amount_msat * settings.lightning_fee_percent / 100.0),
     )
+
+
+def calculate_number_of_blank_outputs(fee_reserve_sat: int):
```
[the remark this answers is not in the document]
xphade: Good point, I will add the type! 

In general, I think there are quite a lot of type hints missing in the code base right now, at least Pylance on VS Code is freaking out for me. :slightly_smiling_face: I would like to clean that up at some point and use types consistently throughout the code, best would be if we even enforce it in CI.

@callebtc Are you open for that? If yes, I'd create an issue for it.

callebtc:
LGTM^2

tests/test_mint.py
xphade: > What I meant: Legacy wallets send 4 outputs, but if our new algo says "we only need 3" then legacy wallets would get an error. 

Ah, I think I get what you mean now. However, I think it was already the case before that less outputs could be sent back (e.g. if the return output fits into 3). See line 557 and following in the old version:

```python
return_amounts = amount_split(user_paid_fee_sat - ln_fee_sat)
# we only need as many outputs as we have change to return
outputs = outputs[: len(return_amounts)]
```

At least if I understand that part correctly. :slightly_smiling_face:

PR #225, opened 2023-05-17
Postgres compat 2 - changes to crud.py fix select proof

trbouma:
For some reason, postgres doesn't return the values in the same way as sqlite to populate the proofs. Instead of SELECT *  made the select more specific: 
```
SELECT id,amount,c,secret,reserved,send_id,time_created,time_reserved from proofs
```
and iterated through the rows to populate the list of proofs instead of using **dict(). A bit verbose, but it works. I tested it back on sqlite and it works fine.

This might be a workaround, but this is what I need for to make it work with postgres.


callebtc:
You accidentally committed many unrelated changes

trbouma:
Sorry about that, @callebtc I will close this PR and open a new one.

trbouma:
closing

Issue #195, opened 2023-05-08
[Mint] Mint gets stuck sometimes


callebtc:
I don't know what the reason is but when it happens, the mint stops responding to any further API calls and needs to be killed with `kill -9`.

Happens rarely but often enough to be an annoyance. Maybe something to do with Fastapi.

callebtc:
I thought the transaction-lock on the db might solve the issue in #217 but it didn't.

callebtc:
Caught the mint at getting stuck with excessive logging

```
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.78 | TRACE | melt called
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.78 | TRACE | attempting lock table proofs_pending
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.78 | TRACE | locked table proofs_pending
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.78 | TRACE | setting proofs pending
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.78 | TRACE | crud: _validate_proofs_pending validating proofs
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.78 | TRACE | crud: _validate_proofs_pending got proofs pending
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.78 | TRACE | crud: _set_proofs_pending setting proof ............... as pending
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.79 | TRACE | set proofs as pending
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.79 | TRACE | Validating proof with keyset I2yN+iRYfkzT.
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.79 | TRACE | Validating proof with keyset I2yN+iRYfkzT.
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.79 | TRACE | Validating proof with keyset I2yN+iRYfkzT.
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.79 | TRACE | Validating proof with keyset I2yN+iRYfkzT.
May 16 13:13:02 caph poetry[2362686]: 2023-05-16 13:13:02.79 | TRACE | Validating proof with keyset I2yN+iRYfkzT.
[...]

PR #229, opened 2023-05-17
LNBits Integration: use aiohttp instead of requests

AngusP:
As the Mint API is async, it should make any upstream requests with an async library instead of a synchronous/blocking one like *requests*.

This replaces uses of `requests` in the LNBits Lightning integration with [`aiohttp`](https://docs.aiohttp.org/en/stable/)

 This may be a cause of #195  -- other sources of blocking may be filesystem I/O to SQLite, and other blocking network requests e.g. to Nostr relays, PostgreSQL


cashu/lightning/lnbits.py
```diff
@@ -60,37 +63,48 @@ async def create_invoice(
...
-            r = self.s.post(url=f"{self.endpoint}/api/v1/payments", json=data)
+            r = await self.s.post(url=f"{self.endpoint}/api/v1/payments", json=data)
+        except:
+            return InvoiceResponse(False, None, None, "Request error.")
+        try:
             r.raise_for_status()
         except:
-            return InvoiceResponse(False, None, None, r.json()["detail"])
+            data = await r.json()
+            return InvoiceResponse(False, None, None, data["detail"])
```
AngusP, line 73: Separated into two different `try/except` blocks as the first can fail because of network problems, the second is for reponse problems.

Also, `data = await r.json()` depends on `r` being a valid response, so we can't access it in the exception handler unless the initial request succeeded to some extent (it may still throw "this isn't JSON!" error if the response was something like a Nginx 502 bad gateway, but that was the behaviour of the original code too)

AngusP:
> Unfortunately, since I regret that we've done duplicate work, I'll have to try to convince you to hop on https://github.com/cashubtc/cashu/pull/230

@callebtc no worries at all, makes sense if `httpx` already in use somewhere 👍

AngusP:
Closing in favour of #230

PR #231, opened 2023-05-18
db: postgres cast int to bigint

### LINKED ISSUE #228: [Mint] DB Compatibility issue - different maximum `INTEGER` sizes, some under theoretical max satoshis (2.1Qn)
In the DB multiple tables use the SQL datatype `INTEGER` to store `amount`s of Satoshis which is sometimes an insufficient size to be able to store possible amounts.

 * :heavy_check_mark:  SQLite `INTEGER` byte-size is dynamic with a [maximum of 8 bytes](https://www.sqlite.org/datatype3.html) so 2^64 bit signed, which can fit 2.1 quadrillion
 * :x: PostgreSQL `INTEGER` [maximum is 4 bytes](https://www.postgresql.org/docs/9.1/datatype-numeric.html) signed which does **not** fit 2.1Qn
 * :heavy_check_mark: Cockroach `INT`/`INTEGER` [is also 8 bytes](https://www.cockroachlabs.com/docs/stable/int.html) so also fits

Note you can hit this condition with a request like [`/mint?amount=210000000000000`](https://legend.lnbits.com/cashu/api/v1/4gr9Xcmz3XEkUNwiBiQGoC/mint?amount=210000000000000) getting back an error when the insert into the `invoices` table fails:

```
{
    "detail": "(psycopg2.errors.NumericValueOutOfRange) integer out of range\n\n[SQL: \n        INSERT INTO cashu.invoices\n          (amount, pr, hash, issued)\n        VALUES (%s, %s, %s, %s)\n        ]\n[parameters: (210000000000000, 'lnbc21000001pjx2njpsp5v3akz4kp9pe4z4taxtdp62njsulfuaw4ckg8j2ezvler3kms6r5qpp59nfvfxrlahpx8sl62c8dtpwam6vmn467lctms3qwads5nfym3u4sdq2vdshx6rdv5xqzjccq ... (8555 characters truncated) ... ltgzm59cqq8zcqqyqqqqlgqqqq0zcqvs9qyysgq4qzwprp4qjl9zralxfpzezlzy9gatzplxf57rzcz8w6x5dqmmlm5tanxq5pzgd7jeg86e5pw0d36yuu9qlavu9u0zrz6q43lqvgw6sgpgzs6du', '2cd2c4987fedc263c3fa560ed585ddde99b9d75efe17b8440eeb6149a49b8f2b', False)]\n(Background on this error at: http://sqlalche.me/e/13/9h9h)"
}
```

(Sidenote that larger amounts like `mint?amount=2100000000000000000` hit a different issue `"CLN method 'invoice' failed with 'msatoshi: should be positive msat or 'any' ..."`)

This PR #226 (or alt #227) checks the amount is 'sane' but can't defend against Postgres being unable to handle a possible (though somewhat ridiculous/reckless) amount of 21.48 whole Bitcoin (`2147483647 + 1` sats).

So either switching all to use `BIGINT` or using that for Postgres only would fix this issue. This problem also becomes more apparent if Cashu ever switches to msat (1/1000th of a sat) as a single sat becomes worth something (:crossed_fingers:!)

callebtc:
Postgres compatibility: Cast columns of type `INT` to `BIGINT`. Closes #228

### DIFF
--- a/cashu/core/db.py
+++ b/cashu/core/db.py
@@ -49,6 +49,12 @@ class Compat:
             return ""
         return "<nothing>"
 
+    @property
+    def big_int(self) -> str:
+        if self.type in {POSTGRES}:
+            return "BIGINT"
+        return "INT"
+
 
 class Connection(Compat):
     def __init__(self, conn: AsyncConnection, txn, typ, name, schema):
--- a/cashu/mint/migrations.py
+++ b/cashu/mint/migrations.py
@@ -16,7 +16,7 @@ async def m001_initial(db: Database):
     await db.execute(
         f"""
             CREATE TABLE IF NOT EXISTS {table_with_schema(db, 'promises')} (
-                amount INTEGER NOT NULL,
+                amount {db.big_int} NOT NULL,
                 B_b TEXT NOT NULL,
                 C_b TEXT NOT NULL,
 
@@ -29,7 +29,7 @@ async def m001_initial(db: Database):
     await db.execute(
         f"""
             CREATE TABLE IF NOT EXISTS {table_with_schema(db, 'proofs_used')} (
-                amount INTEGER NOT NULL,
+                amount {db.big_int} NOT NULL,
                 C TEXT NOT NULL,
                 secret TEXT NOT NULL,
 
@@ -42,7 +42,7 @@ async def m001_initial(db: Database):
     await db.execute(
         f"""
             CREATE TABLE IF NOT EXISTS {table_with_schema(db, 'invoices')} (
-                amount INTEGER NOT NULL,
+                amount {db.big_int} NOT NULL,
                 pr TEXT NOT NULL,
                 hash TEXT NOT NULL,
                 issued BOOL NOT NULL,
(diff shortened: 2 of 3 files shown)


AngusP (approved):
LGTM :+1: 

Question: Are the migrations auto-generated (at-least initially?) by some tool or manual? Wasn't sure so didn't open a PR

AngusP:
May want a migration specifically for PostgreSQL that does a

```sql
ALTER TABLE {table} ALTER COLUMN amount SET DATA TYPE {db.big_int};
```

for each of the tables having an `amount` already

(From [docs](https://www.postgresql.org/docs/current/sql-altertable.html) -- I think the query is right but you'd need to check!)

callebtc:
The migrations aren't generated automatically, we would have to do that to make sure postgres continues to work as expected (or we ignore the issue and pretend as if nobody does anyqay)!

The LNbits integration uses its own Compat class where this is already handled so that should be ok without migration (I assume that's the most widespread Postgres usage with Cashu).

I would ignore and pretend all is good until someone files an issue except if you think otherwise! Would be cleaner to not to but maybe not worth it.

PR #232, opened 2023-05-18
Crud select proof fix

trbouma:
@callebtc - here we go - just the fix for the crud proofs. The code works, but you may want to tighten it up to your own style/standard.


callebtc:
What does this fix? Formatting is off and the explicit column names in the crud seem redundant.

PR #233, opened 2023-05-18
[Wallet] Define responses for API

sihamon:
Define responses of wallet API using pydantic's `BaseModel` and fastapi's `response_model`.


callebtc:
Excellent this should make it a lot easier to follow along and generate better documentation, thank you!

PR #234, opened 2023-05-18
[Mint] LNBits backend refactor httpx

AngusP:
@callebtc tweak for #230 re-using a 'session' for the class rather than creating one for each request


callebtc:
Fantastic, this is the way!

PR #235, opened 2023-05-19
Add secp256k1 deps to dockerfile

ebrakke:
The dockerfile was failing to build because it was missing some dependencies needed by https://github.com/rustyrussell/secp256k1-py

To test:
```
docker build .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,10 @@ FROM python:3.9-slim
 RUN apt-get update
 RUN apt-get install -y curl python3-dev autoconf g++
 RUN apt-get install -y libpq-dev
+
+# Deps for building secp256k1-py
+RUN apt-get install -y build-essential automake pkg-config libtool libffi-dev
+
 RUN curl -sSL https://install.python-poetry.org | python3 -
 ENV PATH="/root/.local/bin:$PATH"
 WORKDIR /app


AngusP:
> failing to build

Interesting, what error were you getting? It built and ran OK for me, albeit using `podman` instead of Docker -- or was there a runtime error?

ebrakke:
It was a build error when trying to do poetry install, saying it couldn’t find make. 

I can try to get the original error message I was seeing, but was unable to build via docker directly on my Mac.

ebrakke:
```
#13 4.378   ChefBuildError
#13 4.378 
#13 4.378   Backend subprocess exited when trying to invoke get_requires_for_build_wheel
#13 4.378   
#13 4.378   'pkg-config' is required to install this package. Please see the README for details.
#13 4.378   
#13 4.378 
#13 4.378   at ~/.local/share/pypoetry/venv/lib/python3.9/site-packages/poetry/installation/chef.py:152 in _prepare
#13 4.386       148│ 
#13 4.386       149│                 error = ChefBuildError("\n\n".join(message_parts))
#13 4.386       150│ 
#13 4.386       151│             if error is not None:
#13 4.386     → 152│                 raise error from None
#13 4.386       153│ 
#13 4.386       154│             return path
#13 4.386       155│ 
#13 4.386       156│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:
#13 4.386 
#13 4.386 Note: This error originates from the build backend, and is likely not a problem with poetry but with secp256k1 (0.14.0) not supporting PEP 517 builds. You can verify this by running 'pip wheel --use-pep517 "secp256k1 (==0.14.0)"'.
#13 4.386 
```

AngusP:
Slightly concerning that his builds OK on Linux but not for you on your Mac... Also for me, `automake` got installed as a dependency from `RUN apt-get install -y curl python3-dev autoconf g++`

Just wondering, is there a chance you have an old version of `python:3.9-slim` on your system? i.e. does the build work without this change if you add `--no-cache` to `docker build ...`?

But otherwise, these dependencies seem safe to add to me :+1:

ebrakke:
No luck even with `--no-cache`. Definitely a mac issue as I can't reproduce on my linux box, but I can reliably repro on two different mac machines. Even deleted all existing python images from both machines to ensure it was pulling the latest.

PR #236, opened 2023-05-23
[Wallet] API: optional amount in `InvoiceResponse`

callebtc:
I don't know why this wasn't caught by mypy.


cashu/wallet/api/responses.py
```diff
@@ -14,23 +14,23 @@ class PayResponse(BaseModel):
 
 
 class InvoiceResponse(BaseModel):
-    amount: int
-    invoice: Union[Invoice, None] = None
-    hash: Union[str, None] = None
+    amount: Optional[int] = None
```
callebtc, line 17: Here is the actual fix, the rest is just `Union[str, None] -> Optional[str]`

PR #247, opened 2023-06-09
Add a GUI to the Python library version 

fusion44:
I want to add a simple GUI to the library using the [Flet](https://flet.dev). The code name is current NutWorks. 

Keep in mind that this gui is pretty much an empty shell at the moment.

## How to run
Run `poetry install` to get the deps then launch the GUI with `poetry run flet cashu/gui/src/main.py`

## Future Plans
The goal is to keep the GUI very simple and just expose the functionality that the CLI provides via a GUI. The GUI will use the same underlying data as the CLI. Changes made via the CLI will be reflected in the GUI after a reload.

### ToDo's
- [x] Implement fetching balances
- [x] Implement sending tokens
- [x] Implement receiving 
- [ ] make the GUI optional via pip. e.g. `pip install cashu[gui]`

 ## Screenshot
image


AngusP:
This is really cool, but I wonder if it'd be better to create a new repo in cashubtc for this GUI, and for it to use [`cashu`](https://pypi.org/project/cashu/) as a library, rather than be part of that lib? I'm just thinking this would keep this repo as the 'core', having only the CLI wallet as a 'barebones' kinda reference implementation, and we have all the other 'nicer' wallets in different repos (like how [eNuts](https://github.com/cashubtc/eNuts) and [cashu.me](https://github.com/cashubtc/cashu.me) already have their own repos) -- we can definitely add one if you'd like.

Also this would avoid potential maintenance issues where say there was difficulty updating/supporting some python package or version here in the 'core' repo because of a sub-dependency of some UI library or other dependency you've got for NutWorks.

callebtc:
> This is really cool, but I wonder if it'd be better to create a new repo in cashubtc for this GUI

Agree with this! The GUI could either import nutshell directly or build on the wallet API that has been implemented since then!

PR #256, opened 2023-06-23
Use m̶u̶l̶t̶i̶p̶r̶o̶c̶e̶s̶s̶i̶n̶g̶ asyncio locks instead of db locks

callebtc:


cashu/mint/ledger.py
```diff
@@ -643,20 +651,17 @@ async def mint(
...
-                    logger.trace(f"invoice paid: {paid}")
-                except Exception as e:
-                    raise e
+
+        if settings.lightning:
+            if not hash:
+                raise Exception("no hash provided.")
+            self.locks[hash] = (
+                self.locks.get(hash) or mp.Lock()
+            )  # create a new lock if it doesn't exist
```
AngusP, line 660: Could make `self.locks` a `defaultdict` so it creates for you?

cashu/mint/ledger.py
```diff
@@ -25,6 +27,9 @@
 
 
 class Ledger:
+    locks: Dict[str, LockBase] = {}  # holds multiprocessing locks
+    proofs_pending_lock: LockBase = mp.Lock()  # holds locks for proofs_pending database
```
AngusP, line 31: Do you know if there's precedent for using `multiprocessing.Lock` with async? Concern here would be the usual that if it's blocking it'll cause similar async problems/thread-lockups to the blocking requests.

There's also `asyncio.Lock` https://docs.python.org/3/library/asyncio-sync.html

callebtc: Thanks for the review! I chose to not use `asyncio.Lock` (I used it before this PR) because eventually we want to be able to run the nutshell mint on multiple workers and only the multiprocessing lock allows you do lock across different threads. See here: https://stackoverflow.com/questions/18213619/sharing-a-lock-between-gunicorn-workers

callebtc: I don't know if there is any interference between the two though, I will give it a test and see if the lock actually works.

AngusP: If it does turn out to be blocking, you can always run in an executor https://docs.python.org/3/library/asyncio-dev.html#running-blocking-code -- adds a bit of overhead but would work

cashu/mint/ledger.py
callebtc: Seems like `defaultdict` doesn't accept classes by default (hehe) and would need some additional boilerplate: https://stackoverflow.com/questions/32932494/defaultdict-of-a-class-get-index-key-of-current-called-instance-inside-a-class

cashu/mint/ledger.py
callebtc: It was indeed blocking. Once two threads ran into the lock, the whole thing stops. I reverted to asyncio locks for now but this won't be usable if we use multiprocessing in the future.

AngusP (approved):
LGTM, assuming you've tested it 😆

cashu/mint/ledger.py
```diff
@@ -643,20 +654,17 @@ async def mint(
...
+        if settings.lightning:
+            if not hash:
+                raise Exception("no hash provided.")
+            self.locks[hash] = (
+                self.locks.get(hash) or asyncio.Lock()
+            )  # create a new lock if it doesn't exist
+            async with self.locks[hash]:
+                async with self.db.connect() as conn:
+                    # will raise an exception if the invoice is not paid
+                    await self._check_lightning_invoice(amount, hash, conn)
```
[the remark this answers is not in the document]
callebtc: I now delete the locks after they're released, I think that should be safe.

AngusP (approved):

PR #270, opened 2023-06-30
[Wallet/mint] P2PK with timelocks

callebtc:
Pay to Public Key (P2PK) allows the sender to lock tokens to a recipient public key. The lock is enforced by the mint and unlocked using a signature of the intended recipient.

## New Secret format
For a normal ecash operation, `Alice` generates a random `secret` that the mint blind-signs. If anyone later presents `secret` and the signature `C` to the mint, the ecash transaction is valid. Naturally, we want to express more complex spending conditions for transactions. That's where P2PK and P2SH spending condition come in. 

To lock tokens to a spending condition, `Alice` needs to generate a special type of `secret`. If the mint encounters a `secret` with a well-known format corresponding to P2PK or P2SH locks, it will demand other conditions to be met in order to validate the transaction. 

### Secret 
The well-known `secret` is as follows:

```json
[
kind <str>, 
  {
    "data": <str>,
    "nonce": <str>,
    "timelock": <int>,
    "tags": [[ "tag_name", "tag_value"],  ... ]
  }
]
```

- `kind` is the kind of lock used, currently: `P2PK` or `P2SH`
- `data` is the lock (pubkey for P2PK, address for P2SH)
- `nonce` is random data (useful if multiple proofs use the same `data` lock)
- `timelock` is the unix time at which the lock will expire (and everyone who knows `secret` can redeem the token)
- `tags` are additional data that are committed to

Here is one example `secret`: 

`["P2PK", {"data": "03e7f4164a8db54e6b6c67bb9c9885d5aec7e24075af8928411478fb9d74aec4ca", "nonce": "dbab884a7b3e427066eb4e0da1776984", "timelock": 1688247220}]`

In the following, we explain how P2PK works and defer documenting P2SH to a later point in time.

## Locking P2PK tokens
`Alice` wants to send `Carol` tokens that only she can unlock. For this, `Carol` provides her public key `<carol_pub>`. `Alice` sends a `POST /split` request to the mint providing `proofs` as inputs and (a subset of) `outputs` as the tokens she wants to send to `Alice`. 

For those tokens, she wants to send to `Carol`, she choses the well-known `secret` format above. She chooses the kind `P2PK`, inserts `<carol_pub>` into the `data` field, and chooses a timelock of 24 hours by settings `timelock` to the current unix timestamp + 24 hours. She proceeds to unblind the signature as she would normally. The resulting tokens are locked by P2PK and only `Carol` can redeem them in the next 24 hours.

## Unlocking P2PK tokens

Spending conditions are enforced by the mint. For `Carol` to unlock proofs in the next `POST /split` request she needs to provide proof that she can indeed unlock the proof (the witness). Note that the witness is revealed to the mint only during the spending of the token, not while locking (since it is blind-signed).

There are two conditions for unlocking a P2PK token that are enforced by the mint.

### Timelock
If the current unix time is later than the timelock specified in the secret, i.e., `now > timelock_int`, the timelock spending path is triggered. There are currently two options to continune. 

- If `secret.tags` contains no `refund` tag, the token is spendable by anyone who knows the secret. 
- If `secret.tags` contains a tag `["refund", <refund_pubkey>`], then the proof can be spent if the owner of `<refund_pubkey>` provided their signature in `proof.p2pksig`. This will typically be the sender's pubkey, in this case `<alice_pub>`.

_Note: This could be extended with a second pubkey of the original sender `Alice`, i.e., a refund pubkey._

### Signature
To redeem the token using the signature path, **`Carol` provides a Schnorr signature on `sha256(secret)`** of each `Proof` she wants to redeem. To do that, Carol redeems the token as usual using the endpoint `POST /split -d <PostSplitRequest>` but she includes a signature `p2pksig` on `sha256(secret)` in every single `Proof` inside the `PostSplitRequest`.

**The `Proof` object is extended by a field `p2pksig`** that is the signature on `sha256(secret)` by `<carol_pub>`. The `p2pksig` is unique to each `Proof` since at least the `nonce` field in `secret` should be unique. A `Proof` is now of the form

```json
{
  "amount": int, 
  "secret": str,
  "C": str,
  "id": str,
  "p2pksig": <signature_hex> <-- new
}
```

### Signature verification
When a token is redeemed (i.e. spent by `Carol`), the mint encounters `secret` and recognizes its well-known format. The mint now demands P2SH validation. For that, the mint first checks whether the time lock is in the past which would validate the spend attempt. If not, the mint parses the `data` field from `secret` which is `<carol_pub>`. The mint then verifies the signature `p2pksig` on `sha256(secret)` is by `<carol_pub>`.

## Implementation details
Nutshell uses `settings.nostr_private_key` to generate the receiving lock (`carol_pub`) and uses it to sign `sha256(secret)`.

_Note: In the current implement, this only works if the nostr private key is stored in hex format. I did not add deserialization of different formats to this part but it's already implemented in the rest of the nostr features._

_Note: the recipient could also be required to sign the entire transaction at once including the outputs but that would require more heavy lifting on the mint side. Please provide feedback._

## Usage
To use this feature in your Nutshell wallet
- add `NOSTR_PRIVATE_KEY` (hex or nsec) to `.env`
- receiver: create a lock with `cashu lock`
- sender: send to this lock with `cashu send <amount> --lock P2PK:<pubkey>`
- receiver: receive with `cashu receive <token>`

### Mint
- Accepts secrets up to 512 characters long
- Refactor P2SH code
- Add P2PK
- Add timelocks check to P2PK and P2SH

### Wallet:
- Remove need to specify lock for `cashu receive` (P2SH and P2PK)
- `cashu lock` now shows P2PK key per default and `cashu lock --p2sh` creates P2SH locks
- The recipient can see their public key lock using the command `cashu lock`. 
-  Recipients of either P2SH and P2PK tokens do not have to enter their lock anymore (`cashu receive <token> --lock <address>` is now `cashu receive <token>`) 
- Add new setting `TIMELOCK_DELTA_SECONDS`
- Fix nostr receive

### Todo
- [x] use Schnorr signatures
- [x] decide on which private key to use
- [x] if nostr: parse nsec from settings, not only hex
- [x] find a better way to serialize the `secret`
- [x] then add optional refund pubkey 
- [x] make timelocks optional as well
- [x] add tests for P2PK
- [x] add tests for timelocks for P2PK and P2SH
- [ ] write NUT (add test vectors)
- [x] add `Secret` object that can be serialized instead of this ugly `:`-seperated string
- [x] sign messages only after hashing

### DIFF
--- a/cashu/core/base.py
+++ b/cashu/core/base.py
@@ -14,9 +14,42 @@ from .p2pk import sign_p2pk_sign
 # ------- PROOFS -------
 
 
+class SecretKind:
+    P2SH = "P2SH"
+    P2PK = "P2PK"
+
+
+class Secret(BaseModel):
+    """Describes spending condition encoded in the secret field of a Proof."""
+
+    kind: str
+    data: str
+    nonce: Union[None, str] = None
+    timelock: Union[None, int] = None
+    tags: Union[None, List[List[str]]] = None
+
+    def serialize(self) -> str:
+        return json.dumps(
+            [
+                self.kind,
+                {
+                    "data": self.data,
+                    "nonce": self.nonce or PrivateKey().serialize()[:32],
+                    "timelock": self.timelock,
+                    "tags": self.tags,
+                },
+            ]
+        )
+
+    @classmethod
+    def deserialize(cls, data: str):
+        kind, kwargs = json.loads(data)
+        return cls(kind=kind, **kwargs)
+
+
 class P2SHScript(BaseModel):
     """
-    Describes spending condition of a Proof
+    Unlocks P2SH spending condition of a Proof
     """
 
     script: str
@@ -36,7 +69,7 @@ class Proof(BaseModel):
     secret: str = ""  # secret or message to be blinded and signed
     C: str = ""  # signature on secret, unblinded by wallet
     p2pksig: Optional[str] = None  # P2PK signature
-    script: Union[P2SHScript, None] = None  # P2SH spending condition
+    p2shscript: Union[P2SHScript, None] = None  # P2SH spending condition
     reserved: Union[
         None, bool
     ] = False  # whether this proof is reserved for sending, used for coin management in the wallet
--- a/cashu/core/base.py
+++ b/cashu/core/base.py
@@ -29,16 +29,16 @@ class Secret(BaseModel):
     tags: Union[None, List[List[str]]] = None
 
     def serialize(self) -> str:
+        data_dict: Dict[str, Any] = {
+            "data": self.data,
+            "nonce": self.nonce or PrivateKey().serialize()[:32],
+        }
+        if self.timelock:
+            data_dict["timelock"] = self.timelock
+        if self.tags:
+            data_dict["tags"] = self.tags
         return json.dumps(
-            [
-                self.kind,
-                {
-                    "data": self.data,
-                    "nonce": self.nonce or PrivateKey().serialize()[:32],
-                    "timelock": self.timelock,
-                    "tags": self.tags,
-                },
-            ]
+            [self.kind, data_dict],
         )
 
     @classmethod
--- a/cashu/core/base.py
+++ b/cashu/core/base.py
@@ -34,6 +34,7 @@ class Proof(BaseModel):
     amount: int = 0
     secret: str = ""  # secret or message to be blinded and signed
     C: str = ""  # signature on secret, unblinded by wallet
+    p2pksig: Optional[str] = None  # P2PK signature
     script: Union[P2SHScript, None] = None  # P2SH spending condition
     reserved: Union[
         None, bool
--- a/cashu/core/base.py
+++ b/cashu/core/base.py
@@ -37,6 +37,11 @@ class Secret(BaseModel):
             data_dict["timelock"] = self.timelock
         if self.tags:
             data_dict["tags"] = self.tags
+        logger.debug(
+            json.dumps(
+                [self.kind, data_dict],
+            )
+        )
         return json.dumps(
             [self.kind, data_dict],
         )
--- a/cashu/core/base.py
+++ b/cashu/core/base.py
@@ -9,6 +9,7 @@ from pydantic import BaseModel
 from .crypto.keys import derive_keys, derive_keyset_id, derive_pubkeys
 from .crypto.secp import PrivateKey, PublicKey
 from .legacy import derive_keys_backwards_compatible_insecure_pre_0_12
+from .p2pk import sign_p2pk_sign
 
 # ------- PROOFS -------
 
@@ -175,6 +176,19 @@ class PostSplitRequest(BaseModel):
     proofs: List[Proof]
     amount: int
     outputs: List[BlindedMessage]
+    signature: Optional[str] = None
+
+    def sign(self, private_key: PrivateKey):
+        """
+        Create a signed split request. The signature is over the `proofs` and `outputs` fields.
+        """
+        # message = json.dumps(self.proofs).encode("utf-8") + json.dumps(
+        #     self.outputs
+        # ).encode("utf-8")
+        message = json.dumps(self.dict(include={"proofs": ..., "outputs": ...})).encode(
+            "utf-8"
+        )
+        self.signature = sign_p2pk_sign(message, private_key)
 
 
 class PostSplitResponse(BaseModel):
--- a/cashu/core/base.py
+++ b/cashu/core/base.py
@@ -19,6 +19,16 @@ class SecretKind:
     P2PK = "P2PK"
 
 
+class Tags(BaseModel):
+    __root__: List[List[str]]
+
+    def get_tag(self, tag_name: str) -> Union[str, None]:
+        for tag in self.__root__:
+            if tag[0] == tag_name:
+                return tag[1]
+        return None
+
+
 class Secret(BaseModel):
     """Describes spending condition encoded in the secret field of a Proof."""
 
@@ -26,7 +36,7 @@ class Secret(BaseModel):
     data: str
     nonce: Union[None, str] = None
     timelock: Union[None, int] = None
-    tags: Union[None, List[List[str]]] = None
+    tags: Union[None, Tags] = None
 
     def serialize(self) -> str:
         data_dict: Dict[str, Any] = {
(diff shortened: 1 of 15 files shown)


cashu/wallet/wallet.py
```diff
@@ -651,24 +656,37 @@ async def mint(
...
+                )
+        # P2PKH
+        elif secret_p2pk_signatures:
+            logger.debug(f"Unlock signature: {secret_p2pk_signatures}")
+            # attach unlock signatures to proofs
+            assert len(proofs) == len(
+                secret_p2pk_signatures
+            ), "wrong number of signatures"
+            for p, s in zip(proofs, secret_p2pk_signatures):
+                p.p2pksig = s
```
callebtc, line 679: Receiver: This part adds the signatures for each token during redemption.

cashu/wallet/wallet.py
```diff
@@ -999,6 +1017,23 @@ async def create_p2sh_lock(self):
...
+    async def sign_p2pk_with_privatekey(self, proofs: List[Proof]) -> List[str]:
+        private_key = PrivateKey(bytes.fromhex(settings.nostr_private_key), raw=True)
+        assert private_key.pubkey
+        return [
+            sign_p2pk_sign(
+                message=proof.secret.encode("utf-8"), private_key=private_key
+            )
+            for proof in proofs
+        ]
+
```
callebtc, line 1036: Receiver: user nostr pubkey as receiving key and private key for signing during redemption.

cashu/wallet/helpers.py
```diff
@@ -174,9 +181,29 @@ async def send(
...
+                    int(
+                        (
+                            datetime.now()
+                            + timedelta(seconds=settings.timelock_delta_seconds)
+                        ).timestamp()
+                    )
+                )
+            )
+            # we can also add some op_return data to the lock just for fun
+            lock += ":rekt"
```
callebtc, line 206: Sender: adds timelock to secret.

cashu/wallet/helpers.py
```diff
@@ -113,20 +119,21 @@ def deserialize_token_from_string(token: str) -> TokenV3:
...
+    if all([p.secret.startswith("P2SH:") for p in proofs]):
+        address_split = proofs[0].secret.split(":")[1]
         p2shscripts = await get_unused_locks(address_split, db=wallet.db)
         assert len(p2shscripts) == 1, Exception("lock not found.")
         script, signature = p2shscripts[0].script, p2shscripts[0].signature
-    else:
-        script, signature = None, None
+    # P2PK signatures
+    elif all([p.secret.startswith("P2PK:") for p in proofs]):
+        p2pk_signatures = await wallet.sign_p2pk_with_privatekey(proofs)
```
callebtc, line 136: Receiver: aggregates all signatures for each `Proof`.

cashu/core/base.py
```diff
@@ -174,6 +176,19 @@ class PostSplitRequest(BaseModel):
...
+    #     """
+    #     Create a signed split request. The signature is over the `proofs` and `outputs` fields.
+    #     """
+    #     # message = json.dumps(self.proofs).encode("utf-8") + json.dumps(
+    #     #     self.outputs
+    #     # ).encode("utf-8")
+    #     message = json.dumps(self.dict(include={"proofs": ..., "outputs": ...})).encode(
+    #         "utf-8"
+    #     )
+    #     self.signature = sign_p2pk_sign(message, private_key)
```
callebtc, line 191: Example code of what it would look like to sign the whole split request (inputs and outputs).

cashu/mint/ledger.py
```diff
@@ -209,35 +212,89 @@ def _verify_proof_bdhke(self, proof: Proof):
...
+
+            # check signature proof.p2pksig against pubkey
+            # we expect the signature to be on the pubkey (=message) itself
+            assert verify_p2pk_signature(
+                message=proof.secret.encode("utf-8"),
+                pubkey=PublicKey(bytes.fromhex(pubkey), raw=True),
+                signature=bytes.fromhex(proof.p2pksig),
+            ), "p2pk signature invalid."
+            logger.trace(proof.p2pksig)
+            logger.trace("p2pk signature valid.")
```
callebtc, line 292: Mint: P2PK verification. If timelock is in the past, the script condition is considered valid. If not, the signature on every token is checked.

cashu/mint/ledger.py
```diff
@@ -209,35 +212,89 @@ def _verify_proof_bdhke(self, proof: Proof):
...
+                # we skip on error for secrets without timelocks
+                logger.error(f"p2sh timelock invalid: {e}")
+                logger.trace(proof.secret)
+
+            # execute and verify P2SH
+            txin_p2sh_address, valid = verify_bitcoin_script(
+                proof.script.script, proof.script.signature
+            )
+            if not valid:
+                raise Exception("script invalid.")
```
callebtc, line 249: Code refactor: P2SH code cleaned up to look like P2PK and timelocks added.

cashu/mint/ledger.py
```diff
@@ -209,35 +214,90 @@ def _verify_proof_bdhke(self, proof: Proof):
...
+            # now we check the signature
+            if not proof.p2pksig:
+                # no signature present although secret indicates one
+                raise Exception("no p2pk signature in proof.")
+
+            # we parse the secret as a P2PK commitment
+            # assert len(proof.secret.split(":")) == 5, "p2pk secret format invalid."
+
+            # check signature proof.p2pksig against pubkey
+            # we expect the signature to be on the pubkey (=message) itself
```
callebtc, line 287: These comments are all wrong and from a previous implementation, please ignore.

cashu/mint/ledger.py
```diff
@@ -209,35 +214,90 @@ def _verify_proof_bdhke(self, proof: Proof):
...
+            assert signature_pubkey, "no signature pubkey present."
+            assert verify_p2pk_signature(
+                message=secret.serialize().encode("utf-8"),
+                pubkey=PublicKey(bytes.fromhex(signature_pubkey), raw=True),
+                signature=bytes.fromhex(proof.p2pksig),
+            ), "p2pk signature invalid."
+            logger.trace(proof.p2pksig)
+            logger.trace("p2pk signature valid.")
+
+            return True
```
callebtc, line 297: P2PK mint validation

cashu/wallet/wallet.py
```diff
@@ -572,13 +565,26 @@ def _meltrequest_include_fields(proofs):
...
+        # temporarily, we use the NostrClient to generate keys
+        try:
+            nostr_pk = NostrClient(
+                private_key=settings.nostr_private_key, connect=False
+            ).private_key
+            self.private_key = (
+                PrivateKey(bytes.fromhex(nostr_pk.hex()), raw=True) or None
+            )
+        except Exception as e:
+            pass
```
callebtc, line 586: Changed to wallet-generated seed in #131

PR #276, opened 2023-07-05
FEAT: improve on workflows and makefile

dni:
- fix, make `test`
- use make command inside workflows
- use caching inside of workflows
- run formatting workflow every time not only on main branch
- update to poetry 1.5.1

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,10 +1,6 @@
 name: checks
 
-on:
-  push:
-    branches: [main]
-  pull_request:
-    branches: [main]
+on: [push, pull_request]
 
 jobs:
   formatting:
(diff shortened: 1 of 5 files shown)


.github/workflows/checks.yml
```diff
@@ -1,52 +1,50 @@
...
-          python-version: ${{ matrix.python-version }}
       - name: Set up Poetry ${{ matrix.poetry-version }}
         uses: abatilo/actions-poetry@v2
         with:
           poetry-version: ${{ matrix.poetry-version }}
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "poetry"
```
AngusP, line 22: Surely you have to set up python before poetry, otherwise what version of python and pip is poetry using? Presumably whichever the system version in the runner image (ubuntu-latest) is, which could break in the future if the versions mismatch enough to cause a problem

callebtc: @dni ?

dni: if you set it up differntly the it is not using the cache though, any ideas? or just scrap the cache?

callebtc:
🚀 @dni

PR #295, opened 2023-07-30
Stamps

callebtc:
Requests mint signature on `C` by DLEQ proof of `C = aY`.

### DIFF
--- a/cashu/core/base.py
+++ b/cashu/core/base.py
@@ -360,6 +360,18 @@ class PostRestoreResponse(BaseModel):
     promises: List[BlindedSignature] = []
 
 
+# ------- API: STAMP -------
+
+class StampSignature(BaseModel):
+    e: str
+    s: str
+
+class PostStampRequest(BaseModel):
+    proofs: List[Proof]
+
+class PostStampResponse(BaseModel):
+    sigs: List[StampSignature]
+
 # ------- KEYSETS -------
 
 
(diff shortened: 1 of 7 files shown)


moonsettler (changes requested):
Bob does not need to know secret_msg to verify kY = C and provide a DLEQ proof verifiable by anyone. This is a conditional proof if secret_msg is not revealed to the third party they can not be sure that the signature is valid even if the DLEQ proof checks out.

While it is trivial to forge (Y, C) type 'signatures', the values are explicitly committed to in the DLEQ here, and thus can be considered valid, if Y = hash_to_curve(x) is satisfied.

```python
def stamp_step1_bob(
    Y: PublicKey, C: PublicKey, a: PrivateKey, p_bytes: bytes = b""
) -> Tuple[PrivateKey, PrivateKey]:
    if p_bytes:
        # deterministic p for testing
        p = PrivateKey(privkey=p_bytes, raw=True)
    else:
        # normally, we generate a random p
        p = PrivateKey()
    assert p.pubkey
    R1: PublicKey = p.pubkey  # R1 = pG
    R2: PublicKey = Y.mult(p)  # type: ignore # R2 = pY
    print(R1.serialize().hex(), R2.serialize().hex())
    e = hash_e(R1, R2, Y, C)
    s = p.tweak_add(a.tweak_mul(e))  # s = p + ea
    spk = PrivateKey(s, raw=True)
    epk = PrivateKey(e, raw=True)
    return epk, spk
```

moonsettler (approved):

callebtc:
> Bob does not need to know secret_msg

Correct! Adjusted appropriately and introduced a new object `ProofY` for that.

PR #296, opened 2023-08-06
[FEAT] Improve tests

dni:
everything works now for me locally.
for description look at my commit messages :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,3 +55,6 @@ build-backend = "poetry.core.masonry.api"
 mint = "cashu.mint.main:main"
 cashu = "cashu.wallet.cli.cli:cli"
 wallet-test = "tests.test_wallet:test"
+
+[tool.pytest.ini_options]
+log_cli = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,3 @@ build-backend = "poetry.core.masonry.api"
 mint = "cashu.mint.main:main"
 cashu = "cashu.wallet.cli.cli:cli"
 wallet-test = "tests.test_wallet:test"
-
-[tool.pytest.ini_options]
-log_cli = true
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -55,3 +55,6 @@ build-backend = "poetry.core.masonry.api"
 mint = "cashu.mint.main:main"
 cashu = "cashu.wallet.cli.cli:cli"
 wallet-test = "tests.test_wallet:test"
+
+[tool.pytest.ini_options]
+log_cli = true
```
callebtc, line 60:
```suggestion
```

tests/conftest.py
```diff
@@ -16,42 +16,36 @@
...
+settings.mint_port = SERVER_PORT
+settings.mint_host = "0.0.0.0"
+settings.mint_listen_port = SERVER_PORT
+settings.mint_url = SERVER_ENDPOINT
+settings.lightning = False
+settings.tor = False
+settings.mint_lightning_backend = "FakeWallet"
+settings.mint_database = "./test_data/test_mint"
+settings.mint_derivation_path = "0/0/0/0"
+settings.mint_private_key = "TEST_PRIVATE_KEY"
```
AngusP, line 33: Hrm, overwriting the settings like this strikes me as a bad idea -- to me intuitively settings should be immutable, avoids having value A at some time, then a bit of code runs, then it's now value B, things break or are confusing. Or potentially if you just import this file, it'll mangle the settings because the mutation doesn't happen in a function. Also as they're currently configured, they'll be type-checked on load (and will error if a setting is wrong), but they won't be when you re-assign them like this.

Instead it'd be 'better' to use a separate `.env` for the tests ([docs](https://docs.pydantic.dev/2.1/usage/pydantic_settings/#dotenv-env-support))

dni: its a hard override on settings for conftest. i think its fine, mixing up different `.env` is error prone too. i could just prefix the poetry run command inside `Makefile` but then running it with `poetry run pytest`  does not work.

PR #297, opened 2023-08-07
[REFACTOR] use bolt11 library

dni:
refactored to use https://pypi.org/project/bolt11/


cashu/mint/ledger.py
```diff
@@ -927,7 +927,8 @@ async def melt(
             logger.trace("verified proofs")
 
             total_provided = sum_proofs(proofs)
-            invoice_obj = bolt11.decode(invoice)
+            invoice_obj = decode(invoice)
```
AngusP, line 930: Minor: would prefer `import bolt11` then `bolt11.decode(invoice)`, here it's a bit of a "huh which decode" when reading (e.g. I'd think of the stdlib `bytes.decode` first)

AngusP (approved):
LGTM (I've not tested though)

cashu/mint/ledger.py
dni: is `from bolt11.decode import decode as bolt11_decode` ok?

AngusP: sure

callebtc (approved):

Issue #286, opened 2023-07-14
[Wallet] Parse URL from tokens correctly


callebtc:
Tokens with a mint URL ending with a trailing `/` are not parsed correctly. Use a real URL parser to avoid these cases:

```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://127.0.0.1:3338//keys
```

Example V3 Token:

```json
{"token":[{"mint":"http://127.0.0.1:3338/","proofs":[{"amount":1,"secret":"Jk1npCuNXsiaMm0gYHfYUreI","C":"03a88b5d215a810795bb943816f75e0ab393ee2442ac03386b1ca622eb3186def3","id":"1cCNIAZ2X/w1"},{"amount":4,"secret":"SINrWa6QoFjxdl3YH0aD6xOy","C":"029511c6cb99f96aef25f38b1e1962d42a54677116e4db11aa3cd3f43c1d6d0e3e","id":"1cCNIAZ2X/w1"},{"amount":16,"secret":"eMzQAPkslA1nOJ6fqZr9aNl4","C":"02c8f97e677b294e231bb95572c48d8c770450b070bb4e84d9ac94afc765f5123c","id":"1cCNIAZ2X/w1"}]}]}
```

```
[...]

PR #298, opened 2023-08-11
Join mint URL and endpoint in a safe way

sihamon:
Solves issue in #286 by joining mint URL and endpoint safely with `urllib.parse.urljoin`.

### DIFF
--- a/cashu/wallet/wallet.py
+++ b/cashu/wallet/wallet.py
@@ -8,6 +8,7 @@ import uuid
 from datetime import datetime, timedelta
 from itertools import groupby
 from typing import Dict, List, Optional, Tuple, Union
+from urllib.parse import urljoin
 
 import requests
 from bip32 import BIP32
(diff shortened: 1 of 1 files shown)


cashu/wallet/wallet.py
```diff
@@ -8,6 +8,7 @@
 from datetime import datetime, timedelta
 from itertools import groupby
 from typing import Dict, List, Optional, Tuple, Union
+from urllib.parse import urljoin
```
AngusP, line 11: `urljoin` can behave a bit weird:

```python
In [1]: from urllib.parse import urljoin

In [2]: urljoin("https://example.com/dir1", "dir2")
Out[2]: 'https://example.com/dir2'  # replaced 🤔 <-- this is the one I think would cause a problem here

In [3]: urljoin("https://example.com/dir1/", "dir2")
Out[3]: 'https://example.com/dir1/dir2'  # expected

In [4]: urljoin("https://example.com/dir1/", "/dir2")
Out[4]: 'https://example.com/dir2'  # maybe expected?

In [5]: urljoin("https://example.com/dir1/", "dir2/")
Out[5]: 'https://example.com/dir1/dir2/'  # expected
```

so this change is OK iff 1) the `url` always has a trailing slash, which I don't think is currently gurarnteed? and 2) the thing being joined never has a leading slash, which is true here but a potential footgun

AngusP (approved):

callebtc (approved):

PR #300, opened 2023-08-17
[DEV] add ruff and remove isort and flake

dni:
- precommit
- workflow
- Makefile

updated black

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -24,9 +24,7 @@ jobs:
         run: poetry install
       - name: Check black
         run: make black-check
-      - name: Check isort
-        run: make isort-check
-  linting:
+  mypy:
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -49,5 +47,8 @@ jobs:
         run: yes | poetry run mypy cashu --install-types || true
       - name: Run mypy
         run: poetry run mypy cashu --ignore-missing
-      - name: Run flake8
-        run: poetry run flake8
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: chartboost/ruff-action@v1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,4 @@
 exclude: '^cashu/nostr/.*'
-
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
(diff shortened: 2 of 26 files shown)


.github/workflows/checks.yml
```diff
@@ -49,5 +47,8 @@ jobs:
         run: yes | poetry run mypy cashu --install-types || true
       - name: Run mypy
         run: poetry run mypy cashu --ignore-missing
-      - name: Run flake8
-        run: poetry run flake8
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: chartboost/ruff-action@v1
```
AngusP, line 54: Neat that the `ruff-action` 'just works', but this isn't necessarily going to be the same version as we'd be using locally with poetry. Also I assume this respects any ruff config in `pyproject.toml` but not sure?

.pre-commit-config.yaml
```diff
@@ -14,16 +13,11 @@ repos:
...
     hooks:
-      - id: isort
-        args: ['--profile', 'black']
-  - repo: https://github.com/pycqa/flake8
-    rev: 6.0.0
-    hooks:
-      - id: flake8
-        entry: poetry run flake8
+      - id: ruff
+        args: [ --fix, --exit-non-zero-on-fix ]
```
AngusP, line 23: For a pre-commit hook, does `--exit-non-zero-on-fix` make sense? If you want to be lazy and let the hook fix for you, it won't work as the non-zero exit will prevent the commit from being written?

cashu/core/base.py
```diff
@@ -161,18 +161,20 @@ class Proof(BaseModel):
 
     id: Union[
         None, str
-    ] = ""  # NOTE: None for backwards compatibility for old clients that do not include the keyset id < 0.3
+    ] = (  # NOTE: None for backwards compatibility for old clients that do not include the keyset id < 0.3
+        ""
+    )
```
AngusP, line 166: Weird but OK, assume `ruff` did this... do you know which rule/style by any chance? Don't think I've seen 'scoped' comments before

AngusP: TBF I'd prefer the comment to move a line up rather than be put in to brackets with the thing it's a comment about, but meh -- i.e. this seems clearer:

```python
# NOTE: None for backwards compatibility for old clients that do not include the keyset id < 0.
id: Union[None, str] = ""
```

pyproject.toml
```diff
@@ -55,3 +54,62 @@ build-backend = "poetry.core.masonry.api"
...
+    ".svn",
+    ".tox",
+    ".venv",
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "venv",
```
AngusP, line 104: Would suggest adding `env` and `.env` too

AngusP (approved):

.github/workflows/checks.yml
dni: yep, it has to respect it, i tried it and else our isort and stuff like that wouldnt validate because its not default. but let see i dont have a lot of experience with it yet.

.pre-commit-config.yaml
dni: good question, i was going of some docs and it works for me as expected, just as black, it fixes it for you automatically

cashu/core/base.py
dni: yep i think those inline same line comments are bad practice in general especially when they are quite long. black cant handle it :).

.pre-commit-config.yaml
dni: i think it just to throw you back and you need to stage the files again and then commit so you see what was done before you actually commit it

callebtc (approved):

PR #302, opened 2023-08-22
[CHORE] updating dependencies, mainly fastapi

dni:
i updated all lib that i could. :) test seem to pass


dni:
it is the "beautiful beast" erroring here with the latest uvicorn package, same in lnbits. do we really need to pass all arguments to uvicorn server? i never used it, tbh. i think its fine to handpick the needed features

PR #306, opened 2023-08-25
check lightning invoice in-flight

### LINKED ISSUE #244: [Nuts] Notes: Async LN payment endpoint design
Notes to myself for implementing async LN payments after we have this nut PR merged: https://github.com/cashubtc/nuts/pull/21

- an async `/melt` endpoint would take in proofs and a LN payment destination
- the endpoint immediately returns a random checking ID (not payment hash!)
- the wallet then repeatedly asks for the payment state including blank outputs for overpaid fees
- if the invoice has been paid, the mint marks it as such and includes the paid fees
- if the invoice is paid, and the wallet asks for the payment state, the mint signs the included blank outputs to return fees
- when the fees are returned, the mint marks the invoice as "fees_returned"

callebtc:
First draft of an async payment flow. Makes use of previous [changes to NUT-07](https://github.com/cashubtc/nuts/pull/21): add new field `pending`.

Closes https://github.com/cashubtc/cashu/issues/244

## Changes
- Adds a new parameter `blocking=False` to the `POST /melt` endpoint which makes it return immediately and pay the invoice in the background.
- The wallet then checks for proof states instead of waiting for the endpoint to return as it did before. 
- If the proof states indicate that the invoice is paid, wallet goes on as usual

## Todo
- [ ] `store_lightning_invoice` only stores the invoice after payment, should store before and update later
- [ ] need to come up with a way to check information about the invoice, otherwise we can't get the preimage of a paid invoice. If there should be a new endpoint to check an invoice state, it should be tied to the proofs used to mint for that invoice, otherwise other users can look up the preimage

### DIFF
--- a/cashu/mint/router.py
+++ b/cashu/mint/router.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import List, Optional, Union
 
 from fastapi import APIRouter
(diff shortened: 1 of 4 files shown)


cashu/mint/router.py
```diff
@@ -184,15 +185,23 @@ async def mint(
...
     logger.trace(f"> POST /melt: {payload}")
-    ok, preimage, change_promises = await ledger.melt(
-        payload.proofs, payload.pr, payload.outputs
-    )
-    resp = GetMeltResponse(paid=ok, preimage=preimage, change=change_promises)
+
+    # run asynchronously if blocking is False
+    if not blocking:
+        asyncio.create_task(ledger.melt(payload.proofs, payload.pr, payload.outputs))
+        resp = GetMeltResponse(paid=False, preimage=None, change=None)
```
AngusP, line 197: Should probably return HTTP 202 Accepted?

edgarmuyomba:
This PR should be closed as the targets were already accomplished

PR #311, opened 2023-09-07
Store proofs used for paying a lightning invoice

### LINKED ISSUE #66: [Mint] Store proofs used for paying a Lightning invoice
That way, a mint can reconstruct a payment in case the user complains about their proofs being stuck as "pending" (for example when the mint crashes at exactly the wrong moment).

sihamon:
Closes #66.

### DIFF
--- a/cashu/mint/crud.py
+++ b/cashu/mint/crud.py
@@ -1,6 +1,7 @@
 from typing import Any, List, Optional
 
 from ..core.base import BlindedSignature, Invoice, MintKeyset, Proof
+from ..core.bolt11 import Invoice as bolt11_invoice
 from ..core.db import Connection, Database, table_with_schema
 
 
@@ -46,6 +47,9 @@ class LedgerCrud:
 
     async def update_lightning_invoice(*args, **kwags):
         return await update_lightning_invoice(*args, **kwags)  # type: ignore
+
+    async def store_invoice_and_proofs(*args, **kwargs):
+        return await store_invoice_and_proofs(*args, **kwargs)
 
 
 async def store_promise(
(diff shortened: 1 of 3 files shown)


cashu/mint/crud.py
```diff
@@ -264,3 +268,28 @@ async def get_keyset(
...
+    conn: Optional[Connection] = None,
+):
+    await (conn or db).execute(
+        f"""
+            INSERT INTO {table_with_schema(db, 'payments')}
+              (invoice_amount, date, payment_hash, preimage, proof_amount, proof_C, proof_secret)
+            VALUES (?, ?, ?, ?, ?, ?, ?)
+            """,
+        (
+            invoice.amount_msat / 1000,
```
AngusP, line 287: Perhaps worth the explicit 'cast' to an int to indicate that's what's going to happen DB-side anyway?
```suggestion
            int(invoice.amount_msat / 1000),
```

cashu/mint/migrations.py
```diff
@@ -156,3 +156,23 @@ async def m007_proofs_and_promises_store_id(db: Database):
...
     )
+
+
+async def m008_paid_invoices_and_proofs(db: Database):
+    """
+    Table to store invoices paid by the mint and corresponding proofs.
+    """
+    await db.execute(f"""
+                CREATE TABLE IF NOT EXISTS {table_with_schema(db, 'payments')} (
+                    invoice_amount {db.big_int} NOT NULL,
```
AngusP, line 167: Storing an amount as a big_int, but it is converted from `msats` to `sats` obove with `/ 1000`, so any sub-sat amount is going to be truncated off... Probably OK, but a potential source of problems?

AngusP (commented):
Sorry not a very useful review 😅 -- the `msat` vs `sat` 'issue' is more general than this PR but seemed worth pointing out anyway

edgarmuyomba:
This appears to have been addressed already — proofs used for melting are now stored in the DB, fully tested too. (verified by manual testing and DB inspection). This can likely be closed.

ye0man:
closing as already resolved

PR #318, opened 2023-09-11
Wallet: Lightning interface

callebtc:
This PR adds a Lightning interface to the wallet, meaning it provides a standard set of functions that can be called by other Lightning apps such as LNbits. 

- Transition from `requests` to `httpx` for Wallet
- Wallet now stures a `mint_id` and `melt_id` for each proof when it is involved in a `/mint` or `/melt`. This allows for the state of the invoice corresponding to the mint/melt to be checked for the tokens.
- Mint/Wallet: rename database columns in table `invoices` (Note: does not work for mint!)
- The `Invoice` object changed: `Invoice.pr` is not `Invoice.bolt11` and `Invoice.hash` is now `Invoice.id` which is the `mint_id` and `melt_id` of the corresponding proofs.
- `FakeWallet` changes: `FakeWallet.get_invoice_state()` returns `paid` with 70% chance, `FakeWallet.pay_invoice` now sleeps for 5 seconds before returning `paid`
- Transitions to using `bolt11` library (closes #297)
- Refactor `ledger.lightning` into own submodule
- Type `LedgerCrud` interface
- Mypy check more strictly 
- Database migrations happen in atomic transactions now

Todo:
- [x] Wallet daemon should startup without mint connection.
- [x] renaming of mint db does not succeed (problem was balance views)
- [x] unset `melt_id` from `proofs` table if the corresponding payment fails
- [x] Write tests for Lightning interface
- [x] Write tests that check that `mint_id` and `melt_id` is correctly set and `melt_id` is unset if payment fails.


cashu/mint/crud.py
```diff
@@ -169,46 +169,49 @@ async def store_lightning_invoice(
...
         WHERE hash = ?
         """,
-        (hash,),
+        (id,),
     )
-
-    return Invoice(**row) if row else None
+    row_dict = dict(row)
+    row_dict["bolt11"] = row_dict["pr"]
+    row_dict["id"] = row_dict["hash"]
```
callebtc, line 200: Could not rename columns?

cashu/mint/migrations.py
```diff
@@ -156,3 +154,20 @@ async def m007_proofs_and_promises_store_id(db: Database):
...
+
+
+#     # rename column pr to bolt11
+#     await db.execute(
+#         f"ALTER TABLE {table_with_schema(db, 'invoices')} RENAME COLUMN pr TO bolt11"
+#     )
+#     # rename column hash to payment_hash
+#     await db.execute(
+#         f"ALTER TABLE {table_with_schema(db, 'invoices')} RENAME COLUMN hash TO id"
+#     )
```
callebtc, line 173: Can't rename columns. SQL logic error.

PR #335, opened 2023-10-08
Add Nix flake (WIP)

jurraca:
This flake exposes a package and a devShell to develop on Cashu and build the binaries.
With flakes enabled, run `nix build` from project root.  
Tested only on Linux. 

currently the cryptography package throws an error on "BN_Add" symbol.


callebtc:
The error you've showed was related to the `bitcoin` library in Cashu which is used for P2SH. I could imagine that we can remove that functionality and with it the entire library from the code base. 

Should I PR that in a separate branch and you can merge it into this branch?

jurraca:
That works yes, thanks!

cashu/wallet/cli/cli.py
```diff
@@ -17,7 +17,7 @@
 from ...core.base import TokenV3
 from ...core.helpers import sum_proofs
 from ...core.settings import settings
-from ...nostr.nostr.client.client import NostrClient
+from nostr.client.client import NostrClient
```
callebtc, line 20: The imports must remain relative otherwise the library can't be imported by other applications. There might also be pip package issues when `nostr` is fetched from its repo.

callebtc:
> That works yes, thanks!

done!

callebtc:
I also moved the nostr client directly into the repo: https://github.com/cashubtc/nutshell/pull/346

jurraca:
OK, updated to use the new nostr setup. Rebased and squashed.
- Not sure if the `poetry.lock` updates here are useful are not, they were automatically generated. 
- This builds the cashu binaries successfully on my machine. Could some mac folk (@chrisguida maybe?) test it as well?
- Starting the mint fails because it's looking for a tor PID in the directory, I'll look at this next.

chrisguida:
I'll get to this in a week or so! Unfortunately traveling without my mac currently

callebtc:
> * Starting the mint fails because it's looking for a tor PID in the directory, I'll look at this next.

Any idea where this comes from? Nothing in the repo knows about PID files afaik....

chrisguida:
just attempted `nix build` on my mac. seems a hash wasn't specified somewhere?

```
> nix build
trace: warning: Unknown cryptography version: '41.0.3'. Please update getCargoHash.
error: hash mismatch in fixed-output derivation '/nix/store/q5zdnqdvji51gldarqrnmvr76q50z8gk-cryptography-41.0.3-vendor.tar.gz.drv':
         specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
            got:    sha256-LQu7waympGUs+CZun2yDQd2gUUAgyisKBG5mddrfSo0=
error: 1 dependencies of derivation '/nix/store/0mqas2hvyns7ipciy0c1mfcs9g4jkskm-python3.10-cryptography-41.0.3.drv' failed to build
error: 1 dependencies of derivation '/nix/store/wxr9qh49r05vhk6mkzm06nb85jqhnqm1-python3.10-cashu-0.14.0.drv' failed to build
```

jurraca:
> Any idea where this comes from?

not sure but it disappeared after rebase so all good

> seems a hash wasn't specified somewhere?

We're avoiding specifying a hash by overriding `cryptography` with `preferWheels = true` so it doesn't try to do a fixed-output derivation. I can reproduce if i don't override the cryptography lib. It's basically [this error](https://github.com/nix-community/poetry2nix/blob/master/docs/edgecases.md#errors-that-are-related-to-rust-and-cargo). Maybe there's an extra mac thing to do?

chrisguida:
Hmm ok, what's the workaround? Should I copy this hash somewhere?

jurraca:
idk tbh. This works on linux, not sure why it fails for mac.

callebtc:
Would love this to work on mac as I'm also one of those people!

callebtc:
Could you guys add a git workflow that checks whether the flake works as the repo might get updated?

PR #342, opened 2023-10-13
Coalesce all witness fields to `Proof.witness`

### LINKED ISSUE #340: [NUTs] P2PK and HTLC – call witness field "witness" only
Store witnesses in only one single field, even if multiple witnesses are provided.

callebtc:
P2PK, P2SH, and HTLC witnesses are now all stored in the same field `Proof.witness`. Further structure inside `Proof.witness` specific to each spending condition type is encoded as JSON.

Closes #340


callebtc:
https://github.com/cashubtc/nuts/pull/40 synched to changes here

PR #389, opened 2024-01-15
Mint: add derivation paths to `.env.example`

callebtc:


trbouma:
Thanks for adding! This had me stumped in the default install and it was not obvious how to set the right unit of account

PR #493, opened 2024-03-25
Issue #313: allow checking pending invoices

### LINKED ISSUE #313: [Freature request] Wallet CLI: allow checking of all pending invoices
- fetch all invoices with `pending == True && out == False` 
- check them all and mint tokens 

with a single command like `cashu invoices -a`.

Guilospanck:
- Updates deprecated `datetime.utcfromtimestamp` in the `cashu/wallet/cli/cli.py` file;
- Adds new options to the `invoices` CLI command:
  - `--only-paid` (`-op`): retrieve only paid invoices;
  - `--only-unpaid` (`-ou`): retrieve only unpaid invoices;
  - `--pending` (`-p`): retrieve only pending invoices.

Pending invoice was defined as invoices with `paid = False` and `out = False` (not sure if this is right).

Closes #313 .

### DIFF
--- a/cashu/mint/conditions.py
+++ b/cashu/mint/conditions.py
@@ -255,9 +255,9 @@ class LedgerSpendingConditions:
         # check if all secrets are P2PK
         # NOTE: This is redundant, because P2PKSecret.from_secret() already checks for the kind
         # Leaving it in for explicitness
-        if not all([
-            SecretKind(secret.kind) == SecretKind.P2PK for secret in p2pk_secrets
-        ]):
+        if not all(
+            [SecretKind(secret.kind) == SecretKind.P2PK for secret in p2pk_secrets]
+        ):
             # not all secrets are P2PK
             return True
 
--- a/cashu/wallet/cli/cli.py
+++ b/cashu/wallet/cli/cli.py
@@ -3,7 +3,7 @@
 import asyncio
 import os
 import time
-from datetime import datetime
+import datetime
 from functools import wraps
 from itertools import groupby, islice
 from operator import itemgetter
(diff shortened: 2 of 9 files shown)


callebtc:
YES! Welcome and thank you.

Guilospanck:
Thanks! Noticed that the checks are not passing. Will try to fix it. I only ran the `poetry run pytest` and thought it was enough.

Guilospanck:
From what I saw the tests are breaking because of the `datetime` (the same reason why `mypy` check is breaking). It was "working on my machine", probably because how it sees the `datetime` module or something like that.

This new commit should work as mypy is also passing.

callebtc:
Were you going to add the "mint tokens" part of issue #313 as well? This only displays the unpaid ones, right?

Guilospanck:
> Were you going to add the "mint tokens" part of issue #313 as well? This only displays the unpaid ones, right?

Yes, I just need more time to dive into the code and understand where and how to do it

Guilospanck:
@callebtc added the _mint token_ part

callebtc:
> @callebtc added the _mint token_ part

Awesome PR thank you! I'd like to add one more nit please, could you change the default behavior to not check (remain offline) and replace the `--test` flag to something like `--check` to do the check?

We'd like to keep the client more offline by default, in case the user just wants to see the invoice requests and not check their state for example.

Guilospanck:
Make sense! Is this it what you meant?

callebtc:
Thank you, this looks great!

I see a bunch of changes in unrelated files. I'm trying to understand why our IDEs seem to have different formatting preferences but this discrepancy doesn't seem to cause a black error in the ruff formatter. Do you have any idea?

Guilospanck:
Good question. I'm not sure. I have the [Black Formatter extension](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter) installed in my vscode, and only these formatters regarding python:

```json
"[python]": {
    "editor.defaultFormatter": "ms-python.black-formatter",
    "editor.codeActionsOnSave": {
      "source.unusedImports": "explicit"
    }
  },
```
Probably something related to [this](https://docs.astral.sh/ruff/faq/#is-the-ruff-linter-compatible-with-black)

PR #516, opened 2024-04-17
Add blink key to startup

findingsov:
I think the blink key is missing here:
for key, value in settings.dict().items():
    if key in [
        "mint_private_key",
        "mint_seed_decryption_key",
        "nostr_private_key",
        "mint_lnbits_key",
        "mint_strike_key",


callebtc (approved):
utACK 

Thank you!

PR #525, opened 2024-05-01
Redirect users to mint home website url if the router path is only "/"

starbackr-dev:
Users typically try to access mint url by putting the mint url on the web browser. e.g. https://mint.minibits.cash/Bitcoin returns this in the browser {"detail":"Not Found"} which is a bad user experience. Instead, if it is mint url without /v1 then redirect the users to a home website page where they can find more info about the mint. e.g. https://stablenut.umint.cash now redirects to https://umint.cash where the user can find more info about the mint. 

This is optional and only works when the env parameter is set.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,6 +42,9 @@ MINT_INFO_MOTD="Message to users"
 
 MINT_PRIVATE_KEY=supersecretprivatekey
 
+# Mint default home website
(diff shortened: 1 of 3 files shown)


callebtc:
This should not be part of a mint but the reverse proxy should handle routing the user to a desired destination.

PR #535, opened 2024-05-16
 add MINT_LND_REST_CERT_VERIFY env bool that when set to False allow to skip certificate validation for LND api call

elliedev80:
- Add MINT_LND_REST_CERT_VERIFY bool variable that when set to False allows verify=False for httpx and ignore LND selfsigned certificate validation

 On branch main
 Your branch is up to date with 'origin/main'.

 Changes to be committed:
	modified:   .env.example
	modified:   cashu/core/settings.py
	modified:   cashu/lightning/lndrest.py

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -69,7 +69,7 @@ MINT_LNBITS_KEY=yourkeyasdasdasd
 MINT_LND_REST_ENDPOINT=https://127.0.0.1:8086
 MINT_LND_REST_CERT="/home/lnd/.lnd/tls.cert"
 MINT_LND_REST_MACAROON="/home/lnd/.lnd/data/chain/bitcoin/regtest/admin.macaroon"
-MINT_LND_REST_CERT_VERIFY=False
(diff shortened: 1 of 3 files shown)


.env.example
```diff
@@ -69,6 +69,7 @@ MINT_LNBITS_KEY=yourkeyasdasdasd
 MINT_LND_REST_ENDPOINT=https://127.0.0.1:8086
 MINT_LND_REST_CERT="/home/lnd/.lnd/tls.cert"
 MINT_LND_REST_MACAROON="/home/lnd/.lnd/data/chain/bitcoin/regtest/admin.macaroon"
+MINT_LND_REST_CERT_VERIFY=False
```
callebtc, line 72:
```suggestion
MINT_LND_REST_CERT_VERIFY=True
```

callebtc: Let's set it to `True`

callebtc (approved):
Thank you, looks good

PR #539, opened 2024-05-31
Add "MINT_ICON_URL" Field for Mint Profile Images

swedishfrenchpress:
I have added the `MINT_ICON_URL` field to allow mint operators to include a profile image. This feature enables applications to index and display the image, making it easier for users to identify the mint.

The Fedimint protocol provides a similar feature with the `federation_icon_url`. Following their naming convention, I propose `MINT_ICON_URL` as the metadata field.

Below is a visual example of how an app might index and display the mint metadata fields. The profile icon can be a useful addition to the existing options.
mint-profile-example

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,7 +39,7 @@ MINT_INFO_DESCRIPTION="The short mint description"
 MINT_INFO_DESCRIPTION_LONG="A long mint description that can be a long piece of text."
 MINT_INFO_CONTACT=[["email","contact@me.com"], ["twitter","@me"], ["nostr",  "npub..."]]
 MINT_INFO_MOTD="Message to users"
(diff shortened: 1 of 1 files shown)


fernandofreamunde:
I can also see a need for a  `MINT_BRANDING` for an optional array of colors, or just one for a possible background color...

can also see that being important for font color, if the mint maker is really into it... but both would end up being optional since the wallet developer will also have a say on how to style their app... but I see it as a good thing...

callebtc:
This is a great idea, but adding this setting here in nutshell won't do anything. We would have to propose a change to NUT-06 so that the mint can communicate this setting at all, so I'd say it's best to start in the specs repository: https://github.com/cashubtc/nuts/blob/main/06.md

Issue #301, opened 2023-08-21
[Wallet] Coinselection algorithm produces too many small coins


callebtc:
The current coinselection algorithm favors large coins first since it assumes that their anon set is small. However, over a long period, this leads to wallets with an ever increasing number of tokens since they tend to be broken down. We need to find a good balance between efficiency and other objectives, such as getting rid of large coins if possible. 

One option would be adding a heuristic threshold for the selection of large coins (i.e., favor selection of coins with a amount if the amount is > 2^14).

sj-fisher:
Thanks for flagging this as a "good first issue", I've been looking for some way to make a small contribution to this interesting project.

I think our (conflicting) objectives for wallet management are:

- Keep transaction messages small by limiting the number of proofs we need to include.
- Keep our wallet small by limiting the number of proofs we need to store. (A single wallet probably has negligible size even on the smallest device, but a hypothetical server-based backup could have large numbers of wallets to store.)
- Help keep the mint database small(er) by limiting the number of proofs we ask it to create.
- Keep the anonymity set for our transactions as large as possible by preferring "common" coin sizes.

Did I miss anything?

If we're really worried about anonymity set size for larger coins, the wallet *could* be anti-social (bloating the mint database) and split such coins as soon as it receives them. If we don't do that, does that mean we're not that bothered about it but we'll take low-cost steps to spend in an anonymity set-friendly way because it doesn't hurt and might help a bit?

Could/should the mint track coin sizes which have very small anonymity sets and try to dynamically phase in or out the use of "large" denominations as demand changes? For example, when generating a new keyset, cap the maximum coin size at something that would have been reasonably commonly issued across the set of mint requests during the lifespan of the previous keyset?

Is it actually a good idea from an anonymity set perspective to spend large coins ASAP? If these coins have a low but non-zero rate of issuance, the longer we hold our large coin, the more other coins of this size are likely to come into existence and the larger the anonymity set grows. (If my thinking is correct here, the anonymity set grows as coins of a particular denomination and keyset are issued. Those coins being spent doesn't shrink the anonymity set, unless there's only ever one such coin in circulation at any time.)

Without having done any simulations or anything like that (yet), my current inclination is to implement coin selection so that:
- If we can pay exactly with a subset of the coins on hand, we do that.
- Otherwise, if we have any coins larger than the required amount, we split the smallest of those coins.
[...]

PR #541, opened 2024-06-05
Implement new coin selection algorithm (#301)

sj-fisher:
This is an attempt at implementing a new coin selection algorithm to address #301. Feedback would be appreciated. Please see issue #301 for some notes on how this works and the thinking behind it.

This is a draft pull request because there is one test failure with this code which I would like some advice on (tests/test_wallet_restore.py:364).


callebtc:
Hi there! As you might have noticed, the coin selection (I should call it _output selection_) algorithm has completely changed a few weeks ago. The wallet now selects outputs so that it has at least `n` of each amount available as a goal. The issue with many small values accumulating should have been fixed with that as well. 

Have you had a chance to look at it yet? Here is the function that selects the outputs for a swap: https://github.com/cashubtc/nutshell/blob/8af1b61b30351550118d87138197a5b63aef0421/cashu/wallet/wallet.py#L377

sj-fisher:
Thanks, I hadn't seen that. Aiming to keep the same number of each denomination of coin in the wallet is a neat idea! Should I close this (and issue #301) or will you?

PR #547, opened 2024-06-15
Fix false mpp payment handling of unsupported backends

callebtc:


lollerfirst (approved):
This will most likely fix the problem.
Unrelated but I noted that `get_payment_quote` could potentially moved to the base class for all back-end since if I am not mistaken it does the same thing. You might want to do it while you're at it.

PR #550, opened 2024-06-16
[Mint]: refactor mint statup connection

### LINKED ISSUE #518: [Mint] Return more human-friendly error if mint can't reach backend during startup
If the connection to the mint and the status check after `cashu.mint.startup:start_mint_init:91 ` fails, the user is greeted with a beautiful stack trace that doesn't tell them much. This exception should be caught and a more friendly error should be returned that explains that the connection to the backend failed.

Dayvvo:
fixes #518

### DIFF
--- a/cashu/mint/startup.py
+++ b/cashu/mint/startup.py
@@ -88,7 +88,10 @@ async def rotate_keys(n_seconds=60):
 
 
 async def start_mint_init():
-    await migrate_databases(ledger.db, migrations)
-    await ledger.startup_ledger()
-    logger.info("Mint started.")
+    try:
+        await migrate_databases(ledger.db, migrations)
+        await ledger.startup_ledger()
+        logger.info("Mint started.")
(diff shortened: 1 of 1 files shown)


cashu/mint/startup.py
```diff
@@ -88,7 +88,10 @@ async def rotate_keys(n_seconds=60):
...
 
 async def start_mint_init():
-    await migrate_databases(ledger.db, migrations)
-    await ledger.startup_ledger()
-    logger.info("Mint started.")
+    try:
+        await migrate_databases(ledger.db, migrations)
+        await ledger.startup_ledger()
+        logger.info("Mint started.")
+    except:
```
callebtc, line 95: Could you also print the Exception `e`, and use `logger.error()`?
```suggestion
    except Exception as e
```

callebtc (changes requested):

Issue #536, opened 2024-05-23
Feature request: add NUT-15 Multinut payments to `CLNRestWallet`


callebtc:
It would be great to have MPP with CLN as well. Here is how it works with LND: https://github.com/cashubtc/nutshell/blob/61cf7def24fa7104d2d195f0cb5105ee8687773f/cashu/lightning/lndrest.py#L198

lollerfirst:
I'm confused: doesn't CLN always use MPP if not specifically disabled in the settings? Maybe you just mean support for partial invoice payments for CLN
Also, doesn't this check prevent partial payments entirely?
https://github.com/cashubtc/nutshell/blob/d80280e35d8947ee866df6c15dec810767bf30b2/cashu/mint/ledger.py#L769-L775

lollerfirst:
I take it that in NUT-15, Alice creates a single bolt11 invoice for a certain amount, which is her amount across all the different mints. So then when she supplies that invoice and the proofs that correspond to a partial amount on a single mint, doesn't this check (NUT-5) prevent partial payments entirely?
https://github.com/cashubtc/nutshell/blob/d80280e35d8947ee866df6c15dec810767bf30b2/cashu/mint/ledger.py#L769-L775

lollerfirst:
Ok I was confused now I managed to understand this anyway, thanks.

PR #551, opened 2024-06-16
Mint: Add `clnrest.py` Lightning backend

lollerfirst:
- Implements changes requested in Issue #536 
- small improvements of the checks on the same functionality in lndrest
- Ports backend for CLN to the clnrest plugin (no more dependency on external c-lightning-REST!)


lollerfirst:
I merged @rbensberg branch which puts corelightningrest.py onto the newer plugin `clnrest` which comes integrated with CLN and should just work without too many hoops.

Although NOW for some reason it hangs when verifying the payment of the quote (when minting tokens). If you stop it and then check the invoice manually `cashu invoice <amount> --id <quote_id>` doesn't have any problem finding it.

lollerfirst:
`paid_invoice_stream` was the problem. Fixed it.

callebtc:
Thank you for the fixes in `lndrest.py`. Could you put those changes into a separate PR?

lollerfirst:
Yea sure.

PR #555, opened 2024-06-20
fixed capitalised typo

0xg4tt0:
curl -ssl https://install.python-poetry.org 

instead of 

curl -sSL https://install.python-poetry.org 

#pedantictester

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ pyenv init
 pyenv install 3.10.4
 
 # install poetry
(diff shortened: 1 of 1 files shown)


0xg4tt0:
note: this edit is only of the README.md

callebtc:
[docs](https://python-poetry.org/docs/#installing-with-the-official-installer) say:
image

PR #561, opened 2024-06-25
[Mint] Add support for BTC and EUR in `StrikeWallet` backend, add EUR to `FakeWallet`

callebtc:
This pull request adds support for EUR in the Strike backend. Previously, only USD was supported. With this change, users will be able to use Strike for BTC, USD, and EUR.

- Configuration in `.env` changes from `StrikeUSDWallet` to `StrikeWallet`
- `FakeWallet` now supports EUR as well
- Invoice listener only for backends that support incoming payment streams
- Invoice listener now restarts on error


daveajones:
Thank you for this. Was just trying to get this working today. 🙏🙏

PR #565, opened 2024-06-30
Mint: add Nostr Wallet Connect as a backend

gudnuf:
This PR allows you to back your mint with any lightning wallet that supports Nostr Wallet Connect.

Adds:
-  `mint_nwc_url` to settings
- `NWCClient` to nostr module
- nwc as lightning backend
- nwc event info to `event.py`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -57,7 +57,7 @@ MINT_DERIVATION_PATH="m/0'/0'/0'"
 MINT_DATABASE=data/mint
 
 # Funding source backends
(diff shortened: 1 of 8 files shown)


gudnuf:
Just realized there should be a [`get_info`](https://github.com/nostr-protocol/nips/blob/master/47.md#get_info) request to make sure the connected NWC supports all methods we need. Will add soon

callebtc:
Really nice work, especially working it into the existing Python nostr client. Thank you.

PR #601, opened 2024-07-26
Mint: LNbitsWallet add extra check for payment state

callebtc:
Needs testing: 
- mint with lnbits backend with lnd backend
- create melt quote
- turn off lnd 
- try melting
- melt should fail


minibits-cash:
PR tested using test mint. My test mint does not use LND but LNbits with Fakewallet.

1. Test with melt while LNbits running - OK
2. Test with melt while LNbits turned off - mint returned error: 

```
return PaymentResponse(error_message=r.json()["detail"])
UnboundLocalError: local variable 'r' referenced before assignment"
```
When looking at the code "r" seems to be defined only inside of try block, so not available in catch block to return proper error message?

PR #613, opened 2024-09-08
NUT-04: add description

callebtc:
Implements https://github.com/cashubtc/nuts/pull/144/


prusnak:
ISTM that this PR does not correctly set `"description": true` for `MintMethodSetting`, but I might be missing something.

prusnak:
> ISTM that this PR does not correctly set `"description": true` for `MintMethodSetting`, but I might be missing something.

fix in https://github.com/cashubtc/nutshell/pull/617

Issue #614, opened 2024-09-08
Melt result state PAID while related lightning payment FAILED


minibits-cash:
I had a long time stuck melt-related lightning payment in IN_FLIGHT status. In spite of the fact that related expiration block heights expired, it did not clear up until I restarted the lnd node. After that, IN_FLIGHT htlc-s were resolved and payment finally timed-out:

`        {
            "payment_hash": "22e9e1334ee66feceed667fcc10d1db60100b059b8bdfb91beacd4aef4d7dace",
            "value": "****",
            "creation_date": "1725151366",
            "fee": "0",
            "payment_preimage": "0000000000000000000000000000000000000000000000000000000000000000",
            "value_sat": "****",
            "value_msat": "****000",
            "payment_request": "******",
            "status": "FAILED",
            "fee_sat": "0",
            "fee_msat": "0",
            "creation_time_ns": "1725151366317091263",
            "htlcs": [],
            "payment_index": "****",
            "failure_reason": "FAILURE_REASON_TIMEOUT"
        },`

However when I checked related meltquote I found, that Nutshell turned the melt quote state to PAID - this is an issue.

I checked the nutshell logs and at that time I found the line, where fee matched the fee_reserve of the payment:

`Sep 08 20:30:05 nutshell poetry[2793]: 2024-09-08 20:30:05.23 | DEBUG | cashu.mint.ledger:melt:920 | Melt – Ok: True: preimage: 0000000000000000000000000000000000000000000000000000000000000000, fee: **** msat`

I use LNBits as a funding source, here are related LNbits logs just after the restart:

```
lnbits poetry[80066]: 2024-09-08 20:30:04.85 | WARNING | didn't receive checking_id from backend, payment may be stuck in database: 22e9e1334ee66feceed667fcc10d1db60100b059b8bdfb91b

lnbits poetry[80066]: 2024-09-08 20:30:05.22 | INFO | 10.7.1.229:34856 - "GET /api/v1/payments/22e9e1334ee66feceed667fcc10d1db60100b059b8bdfb91beacd4aef4d7dace HTTP/1.1" 200

lnbits poetry[80066]: 2024-09-08 20:39:27.55 | WARNING | Deleting outgoing failed payment 22e9e1334ee66feceed667fcc10d1db60100b059b8bdfb91beacd4aef4d7dace: failed
```

callebtc:
Should be fixed

PR #619, opened 2024-09-11
REFACTOR: `paid` flag -> `PaymentResult` object

lollerfirst:
This PR is to deprecate the `paid` flag in `PaymentResponse` and `PaymentStatus` in favor of a more structured `PaymentResult`, that allows for more flexibility.
I have modified the backends to fill in this new field. The main difference is we DO NOT return `SETTLED` when paying an invoice unless we get an explicit confirmation, instead we return `PENDING` and then `melt` notices this and checks the payment status.
I have also adapted the code in `get_mint_quote` and `melt` to rely on the newly added `result` field.

### DIFF
--- a/cashu/lightning/base.py
+++ b/cashu/lightning/base.py
@@ -60,20 +60,44 @@ class PaymentResponse(BaseModel):
     preimage: Optional[str] = None
     error_message: Optional[str] = None
 
+    @property
+    def pending(self) -> bool:
+        return self.result == PaymentResult.PENDING
+    
+    @property
+    def settled(self) -> bool:
+        return self.result == PaymentResult.SETTLED
+
+    @property
+    def failed(self) -> bool:
+        return self.result == PaymentResult.FAILED
+
+    @property
+    def unknown(self) -> bool:
+        return self.result == PaymentResult.UNKNOWN
 
 class PaymentStatus(BaseModel):
     result: PaymentResult
     paid: Optional[bool] = None
     fee: Optional[Amount] = None
     preimage: Optional[str] = None
+    error_message: Optional[str] = None
 
     @property
     def pending(self) -> bool:
-        return self.paid is not True
+        return self.result == PaymentResult.PENDING
+    
+    @property
+    def settled(self) -> bool:
+        return self.result == PaymentResult.SETTLED
 
     @property
     def failed(self) -> bool:
-        return self.paid is False
+        return self.result == PaymentResult.FAILED
+
+    @property
+    def unknown(self) -> bool:
+        return self.result == PaymentResult.UNKNOWN
 
     def __str__(self) -> str:
         if self.result == PaymentResult.SETTLED:
--- a/cashu/lightning/base.py
+++ b/cashu/lightning/base.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from enum import Enum
+from enum import Enum, auto
 from typing import AsyncGenerator, Coroutine, Optional, Union
 
 from pydantic import BaseModel
@@ -13,8 +13,8 @@ from ..core.models import PostMeltQuoteRequest
 
 
 class StatusResponse(BaseModel):
-    error_message: Optional[str]
     balance: Union[int, float]
+    error_message: Optional[str] = None
 
 
 class InvoiceQuoteResponse(BaseModel):
@@ -34,29 +34,19 @@ class InvoiceResponse(BaseModel):
     payment_request: Optional[str] = None
     error_message: Optional[str] = None
 
-class PaymentResult(Enum):
-    FAILED = 0
-    SETTLED = 1
-    PENDING = 2
 
-    UNKNOWN = 3
+class PaymentResult(Enum):
+    SETTLED = auto()
+    FAILED = auto()
+    PENDING = auto()
+    UNKNOWN = auto()
 
     def __str__(self):
         return self.name
 
-    # We assume `None` is `PENDING`
-    @classmethod
-    def from_paid_flag(cls, paid: Optional[bool]):
-        if paid is None:
-            return cls.PENDING
-        elif not paid:
-            return cls.FAILED
-        elif paid:
-            return cls.SETTLED
 
 class PaymentResponse(BaseModel):
     result: PaymentResult
-    ok: Optional[bool] = None  # True: paid, False: failed, None: pending or unknown
     checking_id: Optional[str] = None
     fee: Optional[Amount] = None
     preimage: Optional[str] = None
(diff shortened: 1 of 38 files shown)


cashu/lightning/base.py
```diff
@@ -33,36 +34,79 @@ class InvoiceResponse(BaseModel):
...
+    FAILED = 0
+    SETTLED = 1
+    PENDING = 2
+
+    UNKNOWN = 3
+
+    def __str__(self):
+        return self.name
+
+    # We assume `None` is `FAILED`
```
prusnak, line 47: I think that `None` should be `PENDING`, no?

lollerfirst: Maybe you are right. But we also used `None` to signal what now would be `UNKNOWN`

lollerfirst: Shouldn't matter too much, though. One shouldn't be using that. I only ever created it to use it in `FakeWallet`

prusnak: Then I suggest to change this code to assume that None is PENDING.

That way it is consistent with rest of the old and new code.

lollerfirst:
I am having trouble with LNBits.

The code for `get_invoice_status` reads a response field `paid` that isn't event mentioned in the docs (https://demo.lnbits.com/docs#/Payments/Payment_List_api_v1_payments_get)

cashu/lightning/base.py
```diff
@@ -33,36 +34,81 @@ class InvoiceResponse(BaseModel):
...
+        if paid is None:
+            return cls.PENDING
+        elif not paid:
+            return cls.FAILED
+        elif paid:
+            return cls.SETTLED
 
 class PaymentResponse(BaseModel):
+    result: PaymentResult
     ok: Optional[bool] = None  # True: paid, False: failed, None: pending or unknown
```
callebtc, line 59:
```suggestion
```

I suggest we remove all legacy flags

cashu/lightning/base.py
```diff
@@ -33,36 +34,81 @@ class InvoiceResponse(BaseModel):
...
+
+    # We assume `None` is `PENDING`
+    @classmethod
+    def from_paid_flag(cls, paid: Optional[bool]):
+        if paid is None:
+            return cls.PENDING
+        elif not paid:
+            return cls.FAILED
+        elif paid:
+            return cls.SETTLED
```
callebtc, line 55:
```suggestion
```

We can get rid of the old `paid` flag altogether.

cashu/lightning/base.py
```diff
@@ -33,36 +34,81 @@ class InvoiceResponse(BaseModel):
...
+    def failed(self) -> bool:
+        return self.result == PaymentResult.FAILED
+
+    @property
+    def unknown(self) -> bool:
+        return self.result == PaymentResult.UNKNOWN
 
 class PaymentStatus(BaseModel):
+    result: PaymentResult
     paid: Optional[bool] = None
```
callebtc, line 83:
```suggestion
```

callebtc:
This will be used to fix #614

tests/test_mint_lightning_blink.py
```diff
@@ -135,7 +136,7 @@ async def test_blink_pay_invoice_failure():
         state=MeltQuoteState.unpaid,
     )
     payment = await blink.pay_invoice(quote, 1000)
-    assert not payment.ok
+    assert payment.result != PaymentResult.SETTLED
```
lollerfirst, line 139:
```suggestion
    assert payment.settled
```

tests/test_mint_lightning_blink.py
```diff
@@ -155,7 +156,7 @@ async def test_blink_get_invoice_status():
     }
     respx.post(blink.endpoint).mock(return_value=Response(200, json=mock_response))
     status = await blink.get_invoice_status("123")
-    assert status.paid
+    assert status.result == PaymentResult.SETTLED
```
lollerfirst, line 159:
```suggestion
    assert status.settled
```

tests/test_wallet_regtest_mpp.py
```diff
@@ -59,19 +59,24 @@ async def _mint_pay_mpp(invoice: str, amount: int, proofs: List[Proof]):
...
+        target=partial_pay_real_invoice, args=(invoice_payment_request, 32, 1)
+    )
 
     t1.start()
     t2.start()
     t1.join()
     t2.join()
 
-    assert wallet.balance <= 256 - 32
+    assert wallet.balance == 64
```
lollerfirst, line 79:
```suggestion
    assert wallet.balance == 128 - 32
```
We have 128 in the wallet, but we only pay 32 from the wallet. The other 32 is paid from another node.

tests/test_mint_lightning_blink.py
```diff
@@ -183,7 +184,7 @@ async def test_blink_get_payment_status():
     }
     respx.post(blink.endpoint).mock(return_value=Response(200, json=mock_response))
     status = await blink.get_payment_status(payment_request)
-    assert status.paid
+    assert status.result == PaymentResult.SETTLED
```
lollerfirst, line 187:
```suggestion
    assert status.settled
```

tests/test_mint_lightning_blink.py
```diff
@@ -102,7 +103,7 @@ async def test_blink_pay_invoice():
         state=MeltQuoteState.unpaid,
     )
     payment = await blink.pay_invoice(quote, 1000)
-    assert payment.ok
+    assert payment.result == PaymentResult.SETTLED
```
lollerfirst, line 106:
```suggestion
    assert payment.settled
```

tests/test_mint_init.py
```diff
@@ -376,4 +396,72 @@ async def test_startup_regtest_pending_quote_failure(wallet: Wallet, ledger: Led
...
+
+    # expect that no melt quote is pending
+    melt_quotes = await ledger.crud.get_all_melt_quotes_from_pending_proofs(
+        db=ledger.db
+    )
+    assert not melt_quotes
+
+    # expect that proofs are unspent
+    states = await ledger.db_read.get_proofs_states([p.Y for p in send_proofs])
+    assert all([s.unspent for s in states])
```
lollerfirst, line 464: That's awesome!

cashu/mint/ledger.py
```diff
@@ -755,9 +758,8 @@ async def get_melt_quote(self, quote_id: str) -> MeltQuote:
             status: PaymentStatus = await self.backends[method][
                 unit
             ].get_payment_status(melt_quote.checking_id)
-            if status.paid:
+            if status.result == PaymentResult.SETTLED:
```
lollerfirst, line 761:
```suggestion
            if status.settled:

```

lollerfirst (commented):
Nice! Left some suggestions...

tests/test_mint_init.py
lollerfirst: Do we only check this on startup?

lollerfirst:
I think this is not needed anymore.

PR #623, opened 2024-09-22
chore: run pyupgrade

prusnak:
- use `{...}` instead of `set([...])`
- do not use `class Foo(object):`, just use `class Foo:`
- do not specify default flags (`"r"`) for `open()`


callebtc (approved):

callebtc:
You don't know how much I love your PRs.

PR #627, opened 2024-09-29
chore: modernize f-string usage

prusnak:


cashu/wallet/wallet.py
```diff
@@ -349,9 +348,8 @@ async def load_proofs(self, reload: bool = False, all_keysets=False) -> None:
                 for keyset_id in self.keysets:
                     proofs = await get_proofs(db=self.db, id=keyset_id, conn=conn)
                     self.proofs.extend(proofs)
-        logger.trace(
-            f"Proofs loaded for keysets: {' '.join([k.id + f' ({k.unit})' for k in self.keysets.values()])}"
-        )
+        proofs_str = ' '.join([f"{k.id} ({k.unit})" for k in self.keysets.values()])
+        logger.trace(f"Proofs loaded for keysets: {proofs_str}")
```
callebtc, line 352:
```suggestion
        keysets_str = ' '.join([f"{k.id} ({k.unit})" for k in self.keysets.values()])
        logger.trace(f"Proofs loaded for keysets: {keysets_str}")

```

cashu/nostr/bech32.py
```diff
@@ -73,7 +73,7 @@ def bech32_create_checksum(hrp, data, spec):
 def bech32_encode(hrp, data, spec):
     """Compute a Bech32 string given HRP and data values."""
     combined = data + bech32_create_checksum(hrp, data, spec)
-    return hrp + "1" + "".join([CHARSET[d] for d in combined])
+    return f"{hrp}1{''.join([CHARSET[d] for d in combined])}"
```
callebtc, line 76: In some places, the f-string format is a bit less intuitive

callebtc:
@prusnak thanks! it's hard to see what changed with the force pushes. I like this PR in almost all places but some strings are definitely harder to read as f-strings. Did you consider the comments above or are they "resolved" because of the force-push?

prusnak:
> Did you consider the comments above or are they "resolved" because of the force-push?

Yes, I addressed both comments.

PR #630, opened 2024-10-02
Fix: wallet include fees in swap outputs for inputs of successive melt

callebtc:
For keysets with fees, in cases where a melt requires a swap, the wallet did not include the ecash fees in the outputs generated by that swap.


lollerfirst (approved):
lgtm

Issue #640, opened 2024-10-16
Wallet does not sort amounts during swap


callebtc:
Wallet should order ~inputs~ outputs to all swaps and remember which outputs to keep or send?

callebtc:
Done! Thanks @lollerfirst

PR #648, opened 2024-10-26
[FIX] Wallet sort outputs before swapping

lollerfirst:
From Issue: #640

>Wallet should order inputs to all swaps and remember which outputs to keep or send?

I am sorting the inputs here but I am not sure what do you mean by "remember which outputs to keep or send". Isn't it already doing it?


callebtc:
I was referring to the outputs, not the inputs. The output sorting can tell the mint what portion of the outputs you're going to keep vs send (imagine you create ouputs `[[...1000 sat], [...53 sat]]`, the mint can infer you're probably sending `1000 sat` to someone. We can sort the outputs per amount before a swap, then un-sort them again so we know what part to keep vs to send (should be a fairly simple change).

lollerfirst:
Should be good now

PR #653, opened 2024-10-29
sort split from split_wallet_state

elnosh:
`split_wallet_state` method is returning an unsorted split. Sorting it so that blinded messages in mint request are sorted.


callebtc:
Simple and clean fix, LGTM

PR #670, opened 2024-11-15
Support NUT-XX (signatures on quotes) for mint and wallet side

lollerfirst:
Adding support for NUT-19.
Main changes:

**Common**:
- `MintQuote` has two more optional fields named `privkey` and `pubkey`
- `nut19.py` contains `construct_message`, `sign_mint_quote` and `verify_mint_quote`
- `PostMintQuoteRequest` has one more optional field named `pubkey`
- `PostMintRequest` has one more optional field named `pubkey`

**Mint**:
- If a `PostMintQuoteRequest` contains a `pubkey`, save the public key together with the quote and
    require a BIP340 signature from said public key on the mint request's quote id and blind messages.
- If a `PostMintQuoteRequest` **does not** contain a `pubkey` while Mint is configured to **require** it, reject the request.
- Signal support for NUT-19 and requirement for NUT-19
- created new directory "nuts" with submodule `nut19` and moved submodule `nuts` in it. (fixed all imports)

**Wallet**:
- When requesting a quote -if the Mint supports NUT19- generate an ephemeral keypair and include `pubkey` in the `PostMintRequest`.
- Include `privkey` in the quote when saving it to DB.
- When minting always pull the quote from DB and if the quote has a non-null `key`, sign the quote id and the blinded message with `key`

Implements: https://github.com/cashubtc/nuts/pull/188


cashu/core/base.py
```diff
@@ -409,6 +409,7 @@ class MintQuote(LedgerEvent):
     unit: str
     amount: int
     state: MintQuoteState
+    key: Union[str, None] = None
```
callebtc, line 412: Please rename to `pubkey` so it's a bit clearer

cashu/core/base.py
```diff
@@ -436,10 +437,11 @@ def from_row(cls, row: Row):
             state=MintQuoteState(row["state"]),
             created_time=created_time,
             paid_time=paid_time,
+            key=row["key"],
         )
 
     @classmethod
-    def from_resp_wallet(cls, mint_quote_resp, mint: str, amount: int, unit: str):
+    def from_resp_wallet(cls, mint_quote_resp, mint: str, amount: int, unit: str, key: Optional[str] = None):
```
callebtc, line 444: Same here (`key` -> `pubkey`)

cashu/core/base.py
```diff
@@ -458,6 +460,7 @@ def from_resp_wallet(cls, mint_quote_resp, mint: str, amount: int, unit: str):
             mint=mint,
             expiry=mint_quote_resp.expiry,
             created_time=int(time.time()),
+            key=key,
```
callebtc, line 463: Same here (`key` -> `pubkey`)

cashu/mint/crud.py
```diff
@@ -421,8 +421,8 @@ async def store_mint_quote(
         await (conn or db).execute(
             f"""
             INSERT INTO {db.table_with_schema('mint_quotes')}
-            (quote, method, request, checking_id, unit, amount, paid, issued, state, created_time, paid_time)
-            VALUES (:quote, :method, :request, :checking_id, :unit, :amount, :paid, :issued, :state, :created_time, :paid_time)
+            (quote, method, request, checking_id, unit, amount, paid, issued, state, created_time, paid_time, key)
+            VALUES (:quote, :method, :request, :checking_id, :unit, :amount, :paid, :issued, :state, :created_time, :paid_time, :key)
```
callebtc, line 425: Same here (`key` -> `pubkey`)

cashu/mint/crud.py
```diff
@@ -440,6 +440,7 @@ async def store_mint_quote(
                 "paid_time": db.to_timestamp(
                     db.timestamp_from_seconds(quote.paid_time) or ""
                 ),
+                "key": quote.key or ""
```
callebtc, line 443: Same here (`key` -> `pubkey`)

cashu/mint/ledger.py
```diff
@@ -459,6 +460,7 @@ async def mint_quote(self, quote_request: PostMintQuoteRequest) -> MintQuote:
             state=MintQuoteState.unpaid,
             created_time=int(time.time()),
             expiry=expiry,
+            key=quote_request.pubkey
```
callebtc, line 463: Same here (`key` -> `pubkey`)

cashu/mint/ledger.py
```diff
@@ -518,13 +520,15 @@ async def mint(
...
         outputs: List[BlindedMessage],
         quote_id: str,
+        witness: Optional[str] = None,
     ) -> List[BlindedSignature]:
         """Mints new coins if quote with `quote_id` was paid. Ingest blind messages `outputs` and returns blind signatures `promises`.
 
         Args:
             outputs (List[BlindedMessage]): Outputs (blinded messages) to sign.
             quote_id (str): Mint quote id.
             keyset (Optional[MintKeyset], optional): Keyset to use. If not provided, uses active keyset. Defaults to None.
```
callebtc, line 530:
```suggestion
```

outdated

cashu/mint/ledger.py
```diff
@@ -549,6 +553,10 @@ async def mint(
             raise TransactionError("Mint quote already issued.")
         if not quote.paid:
             raise QuoteNotPaidError()
+        if quote.key and quote.key != "":
```
callebtc, line 556: Same here (`key` -> `pubkey`)

cashu/mint/migrations.py
```diff
@@ -838,3 +838,12 @@ async def m022_quote_set_states_to_values(db: Database):
             await conn.execute(
                 f"UPDATE {db.table_with_schema('mint_quotes')} SET state = '{mint_quote_states.value}' WHERE state = '{mint_quote_states.name}'"
             )
+
+async def m023_add_key_to_mint_quote_table(db: Database):
+    async with db.connect() as conn:
+        await conn.execute(
+            f"""
+                ALTER TABLE {db.table_with_schema("mint_quotes")}
+                ADD COLUMN key TEXT DEFAULT NULL
```
callebtc, line 847: Same here (`key` -> `pubkey`)

cashu/mint/ledger.py
callebtc: `and quote.key != ""` is not needed, string is falsey if empty

cashu/core/crypto/nut19.py
```diff
@@ -0,0 +1,34 @@
+from hashlib import sha256
+from typing import List
+
+from .secp import PrivateKey, PublicKey
+
+from ..base import BlindedMessage
+
+
+def construct_message(quote_id: str, outputs: List[BlindedMessage]) -> bytes:
+    serialized_outputs = bytes.fromhex("".join([o.B_ for o in outputs]))
```
callebtc, line 10:
```suggestion
    serialized_outputs = bytes.fromhex("".join([o.B_ for o in outputs]))
```

This should be utf-8 encoded?

cashu/wallet/cli/cli.py
```diff
@@ -423,7 +423,10 @@ def mint_invoice_callback(msg: JSONRPCNotficationParams):
 
     # user paid invoice before and wants to check the quote id
     elif amount and id:
-        await wallet.mint(amount, split=optional_split, quote_id=id)
+        quote = await get_bolt11_mint_quote(wallet.db, quote=id)
+        if not quote:
+            raise Exception("Quote not found")
+        await wallet.mint(amount, split=optional_split, quote_id=quote.quote, quote_key=quote.key)
```
callebtc, line 429: `quote_key` -> `quote_pubkey`

cashu/wallet/wallet.py
```diff
@@ -424,13 +429,29 @@ async def request_mint(self, amount: int, memo: Optional[str] = None) -> MintQuo
...
 
+    # TODO: generate secret with BIP39 (seed and specific derivation + counter)
+    async def get_quote_ephemeral_keypair(self) -> Tuple[Optional[str], Optional[str]]:
+        """Creates a keypair for a quote IF the mint supports NUT-19
+        """
+        if not self.mint_info:
+            await self.load_mint_info()
+        assert self.mint_info.nuts
+        nut19 = self.mint_info.nuts.get(QUOTE_SIGNATURE_NUT, None)
+        if nut19 and nut19["supported"]:
```
callebtc, line 448: This should be a method of `mint_info`: `supports_quote_pubkey` for example

cashu/wallet/wallet.py
```diff
@@ -481,13 +502,15 @@ async def mint(
         amount: int,
         quote_id: str,
         split: Optional[List[int]] = None,
+        quote_key: Optional[str] = None,
```
callebtc, line 505: `quote_key` -> `quote_privatekey`

callebtc (changes requested):

callebtc:
I think we should move the file `core/crypto/nut19.py` to a new directory `core/nuts` and clean up nut-specific stuff in the future to this path.

cashu/wallet/wallet_deprecated.py
```diff
@@ -261,6 +261,7 @@ async def request_mint_deprecated(self, amount) -> PostMintQuoteResponse:
             paid=False,
             state=MintQuoteState.unpaid.value,
             expiry=decoded_invoice.date + (decoded_invoice.expiry or 0),
+            pubkey=None
```
callebtc, line 264:
```suggestion
```
Optional parameter doesn't need definition

cashu/wallet/wallet.py
```diff
@@ -521,8 +544,13 @@ async def mint(
         await self._check_used_secrets(secrets)
         outputs, rs = self._construct_outputs(amounts, secrets, rs)
 
+        witness: Optional[str] = None
+        if quote_key:
+            privkey = PrivateKey(bytes.fromhex(quote_key), raw=True)
+            witness = nut19.sign_mint_quote(quote_id, outputs, privkey)
```
callebtc, line 550: I would move the `PrivateKey` instantiation into the `nut19` submodule

cashu/wallet/wallet.py
```diff
@@ -424,13 +429,29 @@ async def request_mint(self, amount: int, memo: Optional[str] = None) -> MintQuo
         Returns:
             MintQuote: Mint Quote
         """
-        mint_quote_response = await super().mint_quote(amount, self.unit, memo)
+        privkey, pubkey = await self.get_quote_ephemeral_keypair()
```
callebtc, line 432: Can we move `get_quote_ephemeral_keypair` also to `nut19.py`?

lollerfirst:
> Nice work. One issue here is that `key` is confusing and it seems like you're storing `pubkey` as well as `privatekey` in the same field, depending on whether it is the mint or the wallet.

yea that was the idea. The alternative is keeping the private key separate but still storing it together with the quote in the DB. And then when we request the quote from DB also return the separate private key.

cashu/wallet/wallet_deprecated.py
lollerfirst: mypy was complaining about this because it's defined as optional but doesn't have a default value.

callebtc:
> The alternative is keeping the private key separate but still storing it together with the quote in the DB.

I think it should have two separate columns `privatekey` for the wallet and `pubkey` for the mint, so it's as explicit as possible.

cashu/wallet/wallet.py
lollerfirst: So if we later want to add derivation from seed+counter functionality we might want to keep it inside the wallet class.

lollerfirst:
I moved `nuts.py` to the `nuts` directory as well because I was getting some import errors.

cashu/wallet/wallet.py
lollerfirst: In general depends on the state of the wallet so...

callebtc (approved):

PR #680, opened 2024-12-15
SQLite WAL mode for journaling

callebtc:
Add support for Write-Ahead Logging (WAL) mode in SQLite by executing the appropriate PRAGMA command. This change enhances database performance and concurrency.


a1denvalu3:
I think this can be closed. we already have WAL mode in nutshell

PR #689, opened 2025-01-04
config cleanup regarding MPP

prusnak:
* first two commits are straightforward, they just remove MINT_LND_ENABLE_MPP=True in `.env.example` and `.github/workflows/regtest.yml` since this is default since PR [#667](https://github.com/cashubtc/nutshell/pull/667/commits/457ac30823b3a2f6d2cd448156b8a0fcd0f41867)
* the third commit removes `MINT_CLNREST_ENABLE_MPP: false` from `.github/workflows/regtest.yml` because we probably want to test the default behaviour (which is also True since PR [#667](https://github.com/cashubtc/nutshell/pull/667/commits/457ac30823b3a2f6d2cd448156b8a0fcd0f41867))

Or is there any particular reason why want to keep MPP disabled for CLN in the CI?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -90,10 +90,6 @@ MINT_LND_REST_CERT="/home/lnd/.lnd/tls.cert"
 MINT_LND_REST_MACAROON="/home/lnd/.lnd/data/chain/bitcoin/regtest/admin.macaroon"
 MINT_LND_REST_CERT_VERIFY=True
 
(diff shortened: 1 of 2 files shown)


callebtc:
> Or is there any particular reason why want to keep MPP disabled for CLN in the CI?

Good catch, no we should test it.

PR #690, opened 2025-01-04
update asyncpg 0.29.0 -> 0.30.0 to make Python 3.13 happy

prusnak:
asyncpg 0.29.0 is not compatible with Python 3.13


prusnak:
ping @callebtc

callebtc:
Thank you, I was wondering why 3.13 didn't work

PR #691, opened 2025-01-09
chore: use poetry the latest 1.8.x poetry consistently

prusnak:
there is a new poetry 2.0.0 in the wild and 1.8.5 is the latest of 1.x series

let's stick to that particular version consistently everywhere for now


callebtc:
Awesome, just had this exact issue when doing a docker install recently because the latest poetry CLI flags seem to have changed.

Issue #616, opened 2024-09-10
Mint: Pending payment check on startup should be non-blocking and periodic task


callebtc:
We call `await self._check_pending_proofs_and_melt_quotes()` once during startup in `ledger.py`. This call is blocking but it doesn't need to be, it could run in the background. 

We should also call this function periodically to make sure the mint cleans up any stuck payments that could have been caused due to communication errors with the Lightning backend. This can be solved using an endless task loop that repeats, say, once an hour.

cjbeery24:
PR for fix:
https://github.com/cashubtc/nutshell/pull/707

PR #707, opened 2025-02-20
#616: Fixed pending payment check on startup

cjbeery24:
Removed blocking call from ledger startup. Instead added it to a background task that repeats every hour.


lollerfirst (changes requested):
"hourly" seems pretty arbitrary. would you consider parametrize this and add in the setting?

lollerfirst:
also seems like `test_startup_fakewallet_pending_quote_failure` needs to be adjusted.

Issue #700, opened 2025-02-08
Decode V4 tokens into JSON


patrickReiis:
Is it possible to decode V4 tokens into JSON using cashu CLI?  

I can't find it in the help command:  
```
Usage: cashu [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --host TEXT    Mint URL (default: https://8333.space:3338).
  -w, --wallet TEXT  Wallet name (default: wallet).
  -u, --unit TEXT    Wallet unit (default: sat).
  -d, --daemon       Start server for wallet REST API
  -t, --tests        Run in test mode (don't ask for CLI inputs)
  --help             Show this message and exit.

Commands:
  pay       Pay Lightning invoice.
  invoice   Create Lighting invoice.
  swap      Swap funds between mints.
  balance   Balance.
  send      Send tokens.
  receive   Receive tokens.
  burn      Burn spent tokens.
  pending   Show pending tokens.
  lock      Generate receiving lock.
  locks     Show unused receiving locks.
  invoices  List of all invoices.
  wallets   List of all available wallets.
  info      Information about Cashu wallet.
  restore   Restore backups.
  selfpay   Refresh tokens.
```

lollerfirst:
It's probably a good tool to have

patrickReiis:
Could you do it sir?

lollerfirst:
@patrickReiis

patrickReiis:
Thank you very much!!!  

(when it gets merged I close this issue)

patrickReiis:
Fixed: https://github.com/cashubtc/nutshell/pull/709

PR #709, opened 2025-02-21
Add CLI Decode Command

lollerfirst:
This PR does two things:
* Adds a `cashu decode [OPTIONS] [TOKEN]` command that decodes a token without receiving it and prints it in JSON format.
* In `cashu/wallet/wallet.py:403-405` moves the calls for loading and activating the keyset into the try block so
they don't crash the wallet when the Mint is unreachable.

Related Issue: https://github.com/cashubtc/nutshell/issues/700

### DIFF
--- a/cashu/core/base.py
+++ b/cashu/core/base.py
@@ -922,6 +922,9 @@ class Token(ABC):
     @abstractmethod
     def unit(self, unit: str): ...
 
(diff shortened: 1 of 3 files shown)


patrickReiis:
awck

callebtc (approved):
LGTM

PR #712, opened 2025-02-25
Amountless Melt Requests

lollerfirst:
Adds support for amountless payment requests (e.g. bolt11 amountless invoices)

### DIFF
--- a/cashu/core/models.py
+++ b/cashu/core/models.py
@@ -197,7 +197,7 @@ class PostMeltRequestOptionAmountless(BaseModel):
 
 class PostMeltRequestOptions(BaseModel):
     mpp: Optional[PostMeltRequestOptionMpp]
-    amountless: Optional[PostMeltRequestAmount]
+    amountless: Optional[PostMeltRequestOptionAmountless]
 
 
 class PostMeltQuoteRequest(BaseModel):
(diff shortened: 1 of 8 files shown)


lollerfirst:
It seems like the tests with fakewallet have `nuts = {}` empty nuts dict (?)

callebtc:
Getting back to this very nice PR. 

I think it could benefit from a few more test cases:
- error if an amountless invoice is trying to be paid without providing an amount (should maybe be a mint test, not wallet?)
- and the opposite, error when a normal invoice is being paid but with an amount provided in the request
- tests for the wallet cli for paying amountless invoices

lollerfirst:
> Getting back to this very nice PR.
> 
> I think it could benefit from a few more test cases:
> 
>     * error if an amountless invoice is trying to be paid without providing an amount (should maybe be a mint test, not wallet?)
> 
>     * and the opposite, error when a normal invoice is being paid but with an amount provided in the request
> 
>     * tests for the wallet cli for paying amountless invoices

Yea actually it's testing the Mint, not the wallet. I should probably move that particular test somewhere else.

lollerfirst:
> and the opposite, error when a normal invoice is being paid but with an amount provided in the request

We have MPP quotes remember? A client **can** specify the amount even if the request specifies one.

PR #724, opened 2025-03-10
Update .env.example

findingsov:
should this be 3338?


callebtc:
nope, it's a different server

PR #735, opened 2025-04-19
`SIG_ALL` signature flag for P2PK

callebtc:


cashu/mint/conditions.py
```diff
@@ -252,103 +279,75 @@ def _verify_output_p2pk_spending_conditions(
...
         """
         Verify spending conditions:
-         Condition: P2PK - If sigflag==SIG_ALL in proof.secret, check if outputs contain valid signatures for pubkeys in proof.secret.
+         Condition: If sigflag==SIG_ALL in any proof.secret of the kind P2PK or HTLC
+            we require signatures on all inputs and outputs together.
+
+            Implicitly enforces many other conditions such as all input Secrets
+            being the same except for the nonce (see verify_same_kinds_and_return()).
         """
+        if not self._inputs_require_sigall(proofs):
```
lollerfirst, line 346: Is this the only condition that involves both inputs and outputs? Should be.

lollerfirst (approved):

lollerfirst:
Looks like there is some issue with the public keys that are being passed in.

cashu/mint/conditions.py
callebtc: Yes

PR #737, opened 2025-04-30
human friendly error issue #518

18c83fd3-25ea-4ed9-8205-2abeff9b3883:
Added BackendConnectionError class in cashu/core/errors.py
Updated _check_backends method in cashu/mint/ledger.py to raise the new error
Modified start_mint function in cashu/mint/startup.py to catch and handle the error
Enhanced middleware in cashu/mint/app.py to provide a user-friendly error response
Added test case in tests/test_mint_init.py

### DIFF
--- a/cashu/mint/app.py
+++ b/cashu/mint/app.py
@@ -10,7 +10,7 @@ from fastapi.responses import JSONResponse
 from loguru import logger
 from starlette.requests import Request
 
-from ..core.errors import CashuError
+from ..core.errors import BackendConnectionError, CashuError
 from ..core.logging import configure_logger
 from ..core.settings import settings
 from .auth.router import auth_router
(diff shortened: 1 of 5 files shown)


cashu/mint/app.py
```diff
@@ -29,20 +29,27 @@
...
+                await redis.disconnect()
+                await shutdown_mint()
+            except asyncio.CancelledError:
+                logger.info("CancelledError during shutdown, shutting down forcefully")
+    except BackendConnectionError as e:
+        logger.error(f"Failed to start mint: {e}")
+        # We still need to yield to allow FastAPI to continue
+        # but the app will return errors for all endpoints
+        yield
+        # We don't need to shut down anything as startup failed
```
callebtc, line 52: Do we need to try: except: the entire code block or would it be enough to do it only with `await start_mint()`?

callebtc:
Thank you for this PR. I have only one question, otherwise LGTM!

callebtc:
I've tested it with a failing backend but I was able to curl the keysets endpoint after the startup errored:
```
curl http://localhost:3338/v1/keysets
```

PR #739, opened 2025-05-05
Another changes for LNbits API

iWarpBTC:
Field ["details"]["pending"] is dropped with LNbits v1.0.0. Changing to ["details"]["status"] should work for both old and new versions. 

Also /api/v1/payments/sse endpoint is dropped, so updating paid_invoices_stream. No backward compatibility here though...


lollerfirst:
Can you please link the relevant documentation?

iWarpBTC:
[/api/v1/payments/{payment_hash}](https://demo.lnbits.com/docs#/Payments/api_payment_api_v1_payments__payment_hash__get)
[/api/v1/ws/{item_id}](https://demo.lnbits.com/docs#/Websocket/websocket_update_post_api_v1_ws__item_id__post)

iWarpBTC:
I can't find any relevant...

callebtc:
Thank you for the PR. LNbits tests seem to be failing

iWarpBTC:
Any ideas how to keep both versions of paid_invoices_stream and choose which one to use? "version" in /api/v1/status response?

tsjk:
> Any ideas how to keep both versions of paid_invoices_stream and choose which one to use? "version" in /api/v1/status response? Edit: But it seems we would need an user id for this, not only wallet key...

.env setting?

lollerfirst:
"Fixed" the retro-compatibility of the paid-invoices stream

PR #740, opened 2025-05-08
[FEAT] Add Unit Specific Limits

lollerfirst:
This PR introduces functionality to set unit-specific limits for:
* PEG_INS
* PEG_OUTS
* MAX_BALANCE


callebtc:
This is awesome thank you. Can we use the opportunity to rename these flags to "max mint" and "max melt" instead of peg-in and peg-out? this is old baggage that I wanted to get rid of for a while

callebtc:
Calling `crud.get_balance` should be discouraged because the view takes a very long time to query on sqlite if the database gets moderately large. Instead, since https://github.com/cashubtc/nutshell/pull/705 there is a balance for each keyset that is being tracked which should be pretty much instant to get.

PR #752, opened 2025-05-13
[FEAT] Golomb-Coded Set Filters

lollerfirst:
Adds a `GET v1/filter/{keyset_id}` API endpoint that returns a compact filter for the spent `Y`s belonging to a particular keyset.

Clients can use this filter to "bulk"-check whether ecash notes are spent. Concretely can be used:
1) During recovery, after collecting the blind signatures.
2) Bulk-checking spent notes after a wallet cold-start.


lollerfirst:
There are CI errors that do not appear to be related to the changes I've made.

PR #761, opened 2025-05-27
Fix db migration v27: sqlite doesn't support UPDATE ... FROM

nostrband:


callebtc:
Thank you, we test the migrations with sqlite in the pipeline, could this be related to the sqlite version you're using?

PR #765, opened 2025-06-08
persist the database when launching through docker compose

jooray:
When launching through docker-compose, the database was not persisted in the container, so every container restart meant lost data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,6 @@ The easiest way to get a mint running is through Docker.
 You can build the image yourself by running the following command. Make sure to adjust the environment variables in `docker-compose.yaml`.
 
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -60,6 +60,7 @@ The easiest way to get a mint running is through Docker.
 You can build the image yourself by running the following command. Make sure to adjust the environment variables in `docker-compose.yaml`.
 
 ```bash
+mkdir -p data/mint # for database, will be passed to the container for persistence
```
callebtc: this _might_ not be necessary, afaik docker creates the dirs it needs. do you now?

callebtc:
@jooray can you remove the readme change so we can merge this?

PR #781, opened 2025-09-03
fix: subscription re-init bug

### LINKED ISSUE #780: Websocket quote updates get only sent to the last subscriber
When creating multiple web socket subscriptions for the same quote ID, only the last established subscription receives updates. The initial response is sent for all listeners, but the update is only received on the last subscriber.

Here is client side code to reproduce the issue:

```ts
import { CashuMint, CashuWallet } from "@cashu/cashu-ts";

const mintUrl = "https://nofees.testnut.cashu.space";
const wallet = new CashuWallet(new CashuMint(mintUrl));

const quote = await wallet.createMintQuote(21);

const sub1 = wallet.onMintQuotePaid(
  quote.quote,
  () => {
    console.log("Sub 1 triggered");
  },
  console.error,
);

const sub2 = wallet.onMintQuotePaid(
  quote.quote,
  () => {
    console.log("Sub 2 triggered");
  },
  console.error,
);
```

a1denvalu3:
# Closes #780 

Server-side subscription list was inadvertently reset on every new subscription due to checking the wrong dictionary when initializing the per-filter list

### DIFF
--- a/cashu/mint/events/client.py
+++ b/cashu/mint/events/client.py
@@ -185,10 +185,8 @@ class LedgerEventClientManager:
             raise ValueError("Max subscriptions reached")
 
         for f in filters:
-            if f not in self.subscriptions[kind]:
-                self.subscriptions[kind][f] = []
             logger.debug(f"Adding subscription {subId} for filter {f}")
-            self.subscriptions[kind][f].append(subId)
+            self.subscriptions[kind].setdefault(f, []).append(subId)
             # Initialize the subscription
             asyncio.create_task(self._init_subscription(subId, f, kind))
 


cashu/mint/events/client.py
```diff
@@ -184,13 +184,13 @@ def add_subscription(
...
-        for filter in filters:
-            if filter not in self.subscriptions:
-                self.subscriptions[kind][filter] = []
-            logger.debug(f"Adding subscription {subId} for filter {filter}")
-            self.subscriptions[kind][filter].append(subId)
+        for f in filters:
+            if f not in self.subscriptions[kind]:
+                self.subscriptions[kind][f] = []
+            logger.debug(f"Adding subscription {subId} for filter {f}")
+            self.subscriptions[kind][f].append(subId)
```
callebtc:
```suggestion
            logger.debug(f"Adding subscription {subId} for filter {f}")
            self.subscriptions[kind].setdefault(f, []).append(subId)
```

Issue #731, opened 2025-03-31
Bug: Mark keysets as "deleted" if they disappear from the keysets response


callebtc:
Right now the wallet only adds new keysets to the db when it encounters them (append-only). However, if a keyset is deleted by the mint, the wallet will keep trying to use it. We should mark keysets as "deleted" if they disappear form the keysets response and not load them to memory.

okekefrancis112:
please can i take this issue

callebtc:
> please can i take this issue

Wish I saw this earlier. If you're still up for it, please do!

kilnhast:
@callebtc Can I work in this issue?

KvngMikey:
hi @callebtc @gudnuf Please can this be assigned to me ?

PR #784, opened 2025-09-10
fix(wallet): mark keysets inactive

TheRealCheebs:
- update load_mint_keysets to mark missing keysets from the mint as inactive in the database.
- adds test test_wallet_marks_deleted_keyset

This should address https://github.com/cashubtc/nutshell/issues/731


TheRealCheebs:
I think I need to run this more before declaring it done. Closing the pr for now.

Issue #671, opened 2024-11-15
Feature request: `WALLET_MAX_FEE_PPK` flag to avoid overpaying fees


callebtc:
The wallet should have a way to reject all swaps that have a fee over a certain threshold.

PR #786, opened 2025-09-15
#Issue 671: Feature request: WALLET_MAX_FEE_PPK flag to avoid overpaying fees

aaronmcdaid:
This is a draft PR to implement #671 is _Feature request: WALLET_MAX_FEE_PPK flag to avoid overpaying fees_

I wanted to add more tests, but I ran into problems with the fact that the test mint's keysets have `input_fee_ppk=0`

There is only one test, which is currently `skipped` by default. To really test this feature properly, we need the mint to have keysets with non-zero `input_fee_ppk`. If you set `tests/conf_test.py:settings.mint_input_fee_ppk = 100`, then this test runs successfully:

`poetry run pytest tests/wallet/test_wallet_cli.py::test_send_with_input_fee_limit_cli`

However, that settings change then breaks a lot of other tests, as those tests assume no fees. So that's  why I have left the setting at zero in `conf_test.py` and put this skip on the test: `@pytest.mark.skipif(settings.mint_input_fee_ppk != 100, ...`

I (along with _Claude Code_) tried lots of other things to try to get a temporary non-zero-fee keyset, such as rotating the keyset, but I couldn't get it to work


cashu/core/settings.py
```diff
@@ -20,6 +20,15 @@ def find_env_file():
...
         env_file = ""
+
+    # Apply override environment variables at the end
+    # This allows OVERRIDE_* env vars to override .env file settings
+    for key, value in os.environ.items():
+        if key.startswith("OVERRIDE_"):
+            # Remove OVERRIDE_ prefix and set the environment variable
+            actual_key = key[9:]  # Remove "OVERRIDE_" (9 characters)
+            os.environ[actual_key] = value
+
```
callebtc, line 31: Was this the only way you could run a mint with fees to test the wallet?

tests/conftest.py
```diff
@@ -150,3 +151,77 @@ def mint():
 
     yield server
     server.stop()
+
+
+@pytest.fixture(scope="session")
```
callebtc, line 156: Was this necessary to add?

cashu/core/settings.py
aaronmcdaid: I must admit I was hacking a lot of different things, and it's very possible that there is a simpler solution that I missed.

> Was this the only way you could run a mint with fees to test the wallet?

Now that I look at bit more at the other tests, maybe I could copy a test from `tests/mint/test_mint_fees.py` and use that instead. For example, I guess I could make a test that uses this Wallet and Ledger approach.

I''ll try that this evening, if I have time

image

aaronmcdaid: *PS*: Partial good news. Just a moment ago, I pushed a single commit with a simple new test. If the `max_input_fee_ppk` argument is too low then, as expected, it raises an Exception to complain about the fees being too high

But, when the `max_input_fee_ppk` argument is large enough then it _should_ success but instead generates a strange exception: `FAILED tests/wallet/test_wallet_fees.py::test_send_with_input_fee_limit - Exception: Mint Error: inputs (64) - fees (0) vs outputs (59) are not balanced. (Code: 11000)
`

The fees are 5 sats, so why does this error say 0 sats?

If this can be resolved, then I think the complex/ugly second mint can be removed from this PR

cashu/wallet/cli/cli.py
```diff
@@ -276,11 +288,21 @@ async def pay(
     if wallet.available_balance < total_amount + ecash_fees:
         print(" Error: Balance too low.")
         return
+    from ..errors import InputFeeExceedsLimitError
```
a1denvalu3, line 291: maybe move this up?

cashu/wallet/cli/cli.py
```diff
@@ -528,9 +556,19 @@ async def swap(ctx: Context):
     total_amount = melt_quote.amount + melt_quote.fee_reserve
     if outgoing_wallet.available_balance < total_amount:
         raise Exception("balance too low")
-    send_proofs, fees = await outgoing_wallet.select_to_send(
-        outgoing_wallet.proofs, total_amount, set_reserved=True
-    )
+
+    from ..errors import InputFeeExceedsLimitError
```
a1denvalu3, line 560: same here

cashu/wallet/cli/cli.py
```diff
@@ -687,25 +731,35 @@ async def send_command(
     yes: bool,
     offline: bool,
     include_fees: bool,
+    max_input_fee_ppk: Optional[int],
     force_swap: bool,
 ):
     wallet: Wallet = ctx.obj["WALLET"]
+    from ..errors import InputFeeExceedsLimitError
```
a1denvalu3, line 738: Let's move all of these in one spot at the top

cashu/wallet/cli/cli.py
```diff
@@ -1247,20 +1301,38 @@ async def restore(ctx: Context, to: int, batch: int):
...
     await wallet.load_mint()
 
     # get balance on this mint
     mint_balance_dict = await wallet.balance_per_minturl()
     mint_balance = int(mint_balance_dict[wallet.url]["available"])
     # send balance once to mark as reserved
-    await wallet.select_to_send(
-        wallet.proofs, mint_balance, set_reserved=True, include_fees=False
-    )
+    from ..errors import InputFeeExceedsLimitError
```
a1denvalu3, line 1323: .

PR #797, opened 2025-10-05
Mark Keysets as Deleted.

### LINKED ISSUE #731: Bug: Mark keysets as "deleted" if they disappear from the keysets response

KvngMikey:
Fixes #731 

### Summary
This PR fixes the wallet behavior when a keyset disappears from a mint's `/keysets` response.

### Changes
- Updated `load_mint_keysets` to mark missing keysets as deleted.
- Added `update_keyset_active` in `crud.py` to use named SQL parameters for SQLAlchemy async compatibility.
- Verified with `pytest tests/wallet/test_wallet.py -k keyset` (all keyset tests now pass).


cashu/wallet/wallet.py
```diff
@@ -297,9 +298,19 @@ async def load_mint_keysets(self, force_old_keysets=False):
         logger.trace("Loading mint keysets.")
         mint_keysets_resp = await self._get_keysets()
         mint_keysets_dict = {k.id: k for k in mint_keysets_resp}
-        # load all keysets of thisd mint from the db
+
+        # load all keysets of this mint from the db
         keysets_in_db = await get_keysets(mint_url=self.url, db=self.db)
 
+        # NEW: mark keysets that vanished from mint as inactive
```
callebtc:
```suggestion
        # mark keysets that disappeared from mint as inactive

```

cashu/wallet/wallet.py
```diff
@@ -349,7 +360,14 @@ async def load_mint_keysets(self, force_old_keysets=False):
 
         await self.inactivate_base64_keysets(force_old_keysets)
 
-        await self.load_keysets_from_db()
```
callebtc, line 362: You're only updating the active state in the db but you're not reloading the keysets from the db?

tests/wallet/test_wallet.py
```diff
@@ -553,3 +555,45 @@ async def testactivate_keyset_specific_keyset(wallet1: Wallet):
...
+
+
+@pytest.mark.asyncio
+async def test_keyset_disappears_from_mint(wallet1: Wallet):
+    """
+    Ensure that when a mint no longer returns a previously known keyset,
+    it gets removed from the wallet's keyset list.
+    """
+
+    # ✅ Manually seed the wallet (avoids 'seed not set')
```
callebtc:
```suggestion
    # Manually seed the wallet (avoids 'seed not set')

```

tests/wallet/test_wallet.py
```diff
@@ -553,3 +555,45 @@ async def testactivate_keyset_specific_keyset(wallet1: Wallet):
...
+    mnemo = Mnemonic("english")
+    wallet1.mnemonic = mnemo.generate(strength=128)
+
+    # create a dummy seed from mnemonic (bytes)
+    wallet1._seed = wallet1.mnemonic.encode("utf-8")  # pylint: disable=protected-access
+
+    # ✅ Initialize private key using this dummy seed
```
callebtc: Please remove the remaining AI slop (emojis)

callebtc (changes requested):
LGTM but one outstanding question regarding the removed line `await self.load_keysets_from_db()`

what's the justification for that?

cashu/wallet/wallet.py
KvngMikey: Good catch, I've just added a re-fetch right after marking key-sets inactive so self.keysets immediately hides the vanished ones. Test still passes and the wallet no longer sees inactive key-sets in memory.

a1denvalu3 (approved):

callebtc:
> question: when this PR goes in, would we need to update NUT01 to show that keysets can now be marked as deleted ?
> 
> cc @callebtc @a1denvalu3

nope, this is just for the wallet and not related to the API. the mint just deletes the keyset

cashu/core/base.py
```diff
@@ -779,6 +782,7 @@ def deserialize(serialized: str) -> Dict[int, PublicKey]:
             valid_to=row["valid_to"],
             first_seen=row["first_seen"],
             active=row["active"],
+            deleted=dict(row).get("deleted", False), # use existing value, or false if column doesn't exist of older DB.
```
callebtc: why is `.get()` needed here?

callebtc: should be unnecessary

cashu/core/base.py
```diff
@@ -716,6 +716,7 @@ class WalletKeyset:
     valid_to: Union[str, None] = None
     first_seen: Union[str, None] = None
     active: Union[bool, None] = True
+    deleted: bool = False
```
callebtc: can we use a `deleted_at` as a timestamp?

a1denvalu3 (approved):

cashu/wallet/wallet.py
```diff
@@ -352,8 +366,15 @@ async def load_mint_keysets(self, force_old_keysets=False):
...
 
+        # RE-FETCH after all DB updates using the centralized method
+        self.keysets = {}
         await self.load_keysets_from_db()
 
+        # for minting/swapping, only keep active keysets,
+        self.keysets = {
+            k_id: k for k_id, k in self.keysets.items() if k.active
+        }
+
```
callebtc: why do we do this?

cashu/wallet/wallet.py
```diff
@@ -352,8 +366,15 @@ async def load_mint_keysets(self, force_old_keysets=False):
 
         await self.inactivate_base64_keysets(force_old_keysets)
 
+        # RE-FETCH after all DB updates using the centralized method
+        self.keysets = {}
```
callebtc: side effect

callebtc (changes requested):

cashu/wallet/wallet.py
KvngMikey: removed.

cashu/wallet/wallet.py
KvngMikey: this has also been removed.

cashu/wallet/wallet.py
```diff
@@ -436,15 +450,16 @@ async def load_proofs(self, reload: bool = False, all_keysets=False) -> None:
...
+        """
         # so that the caller can set unit = None, otherwise use defaults
         if unit == "":
             unit = self.unit.name
         if url == "":
             url = self.url
         keysets = await get_keysets(mint_url=url, unit=unit, db=self.db)
-        for keyset in keysets:
-            self.keysets[keyset.id] = keyset
+        self.keysets = {k.id: k for k in keysets if k.deleted_at is None}
```
callebtc: instead of this, we can add `AND deleted_at = None` for the `get_keysets` crud, what do you think? that eliminates the need to remember doing this after each get_keysets.

KvngMikey: good idea, i'll do this now so we filter at DB level and update the PR.

a1denvalu3:
---
title: "Wallet permanently loses unspent funds when a mint stops reporting a keyset"
slug: wallet-loses-funds-on-deleted-keyset
date: 2026-04-14
status: confirmed
severity: high
target: nutshell
nuts: [NUT-02]
---

## Summary
The changes introduced in PR #797 mark keysets as deleted (`deleted_at = int(time.time())`) if they disappear from the mint's `/v1/keysets` response. However, `get_keysets` defaults to excluding deleted keysets, which causes `load_keysets_from_db()` to remove these keysets from `self.keysets`. Consequently, `load_proofs()` completely ignores any unspent proofs belonging to deleted keysets. This causes valid, unspent user funds to silently vanish from the wallet's balance and memory, rendering them permanently unspendable.

## Root Cause
In `cashu/wallet/wallet.py` (`load_mint_keysets`), keysets missing from the mint's response are marked as deleted:
```python
        for key_id in (local_ids - active_ids):
            ks = keysets_in_db_dict[key_id]
            if ks.deleted_at is None:
                ks.deleted_at = int(time.time())
            await update_keyset(keyset=ks, db=self.db)
```
Subsequently, `load_keysets_from_db()` filters out these keysets. In `cashu/wallet/wallet.py` (`load_proofs`), the wallet only loads proofs if their keyset ID exists in `self.keysets`:
```python
            else:
                for keyset_id in self.keysets:
                    proofs = await get_proofs(db=self.db, id=keyset_id, conn=conn)
                    self.proofs.extend(proofs)
```
Because the deleted keysets are no longer in `self.keysets`, their unspent proofs are skipped entirely.

## Attack Steps
1. A user holds unspent proofs associated with a specific keyset in their wallet.
2. The mint rotates its keysets and decides to remove the old keyset from its `/v1/keysets` payload (e.g., to reduce payload size or deprecate the keyset).
3. The user connects their wallet to the mint, triggering `load_mint_keysets()`.
4. The wallet marks the old keyset as deleted in its database since it is no longer reported by the mint.
5. The wallet calls `load_proofs()`, which fails to load any unspent proofs belonging to the deleted keyset.
6. The user's balance for those proofs drops to zero, and the funds are permanently lost from the wallet UI and internal memory.

## Impact
Any user holding tokens from a keyset that a mint stops returning will permanently lose access to those funds in their wallet. The proofs are silently hidden and cannot be swapped, sent, or recovered without manually editing the SQLite database. This constitutes a high-severity loss of funds/denial of service.

## Test Results
A test script confirming this behavior failed initially (showing the balance dropped from 100 to 0). Applying the proposed fix successfully resolved the test.
```
Balance before mint drop: 100
Balance after mint drop: 0
VULNERABILITY CONFIRMED: Proofs from deleted keyset are no longer loaded into the wallet, causing permanent loss of funds.
```

## Proposed Fix
Before marking a keyset as deleted in `load_mint_keysets`, check if there are any unspent proofs associated with it. If there are, do not mark it as deleted.

```python
        for key_id in (local_ids - active_ids):
            ks = keysets_in_db_dict[key_id]
            
            # Check if there are any unspent proofs for this keyset
            unspent = await get_proofs(db=self.db, id=key_id)
            if unspent:
                logger.warning(f"Keyset {key_id} disappeared from mint but has unspent proofs. Not deleting.")
                continue

            logger.debug(f"Keyset {key_id} no longer reported by mint, marking as deleted")
            if ks.deleted_at is None:
                ks.deleted_at = int(time.time())
            await update_keyset(keyset=ks, db=self.db)
```

a1denvalu3:
---
title: "Reappearing keysets are never undeleted, causing permanent loss of funds"
slug: permanent-keyset-deletion
date: 2026-04-13
status: confirmed
severity: high
target: nutshell
nuts: [NUT-02]
---

## Summary
The wallet tracks keysets that are reported by the mint and marks them as deleted (`deleted_at = int(time.time())`) if they disappear from the mint's active keysets list. However, if a keyset reappears (e.g., after a mint configuration fix, network error, or pagination bug), the wallet fails to clear the `deleted_at` flag. Consequently, the keyset remains permanently excluded from `self.keysets`, rendering any proofs associated with it permanently unspendable by `active_proofs()`.

## Root Cause
In `cashu/wallet/wallet.py` (inside `load_mint_keysets`), when updating existing keysets, the `deleted_at` column is never reset to `None` if the keyset is found in the current mint's response.
```python
        for mint_keyset in mint_keysets_dict.values():
            if mint_keyset.id in keysets_in_db_dict:
                changed = False
                if (not mint_keyset.active ... ):
                    ...
                if (mint_keyset.input_fee_ppk ... ):
                    ...
                # MISSING: Check and reset deleted_at
```
Because `get_keysets` defaults to `exclude_deleted=True`, `load_keysets_from_db()` subsequently fails to load this keyset into `self.keysets`. Thus, `active_proofs()` will permanently filter out any proofs matching this keyset ID, causing a loss of funds for the user.

## Attack Steps
1. The user's wallet loads a valid keyset and mints proofs against it.
2. The mint temporarily fails to include the keyset in its `/v1/keys` response (e.g., due to a temporary misconfiguration, outage, or reverse proxy bug).
3. The user's wallet calls `load_mint_keysets()` and permanently marks the keyset with `deleted_at = timestamp`.
4. The mint resolves the issue, and the keyset reappears in `/v1/keys`.
5. The user's wallet calls `load_mint_keysets()` again. It detects the keyset in the DB and checks for `active` or `fee` changes but fails to reset `deleted_at = None`.
6. The keyset remains deleted in the DB. The user cannot spend any proofs from this keyset.

## Impact
Users can permanently lose access to their funds if the mint temporarily fails to report a keyset. The wallet will permanently drop the keyset, rendering valid proofs unspendable. This constitutes a high-severity denial of service and potential loss of funds.

## Test Results
A test script confirming this behavior failed initially (showing the keyset was permanently excluded from `wallet.keysets`). Applying the proposed fix successfully resolved the test.

## Proposed Fix
In `cashu/wallet/wallet.py`, `load_mint_keysets`, reset `deleted_at` to `None` when an existing keyset reappears in the mint's keyset list:

```python
                if keysets_in_db_dict[mint_keyset.id].deleted_at is not None:
                    keysets_in_db_dict[mint_keyset.id].deleted_at = None
                    changed = True
```

a1denvalu3 (approved):

a1denvalu3:
## Summary
The pull request introduces a `deleted_at` column for keysets, which marks them as deleted if the mint no longer advertises them. However, when redeeming tokens (e.g. during a `cashu receive`), the wallet must verify the DLEQ proofs of incoming tokens using `verify_proofs_dleq`. This verification strictly requires the incoming proofs' keyset ID to be present in memory (`self.keysets`). Since the PR modifies `load_keysets_from_db` to implicitly exclude deleted keysets by default, `verify_proofs_dleq` will raise an `AssertionError` for any valid proof belonging to a deleted keyset, preventing users from receiving valid tokens that the mint would otherwise still honor.

## Root Cause
In `cashu/wallet/wallet.py`, `load_keysets_from_db` populates `self.keysets` from the database:
```python
    async def load_keysets_from_db(
        self, url: Union[str, None] = "", unit: Union[str, None] = ""
    ):
        ...
        keysets = await get_keysets(mint_url=url, unit=unit, db=self.db)
        self.keysets = {k.id: k for k in keysets}
```
The `get_keysets` query defaults to `exclude_deleted=True` (added in `cashu/wallet/crud.py`). This strictly omits keysets with `deleted_at IS NOT NULL`.

During token reception, `redeem()` calls `verify_proofs_dleq()` which verifies the DLEQ of incoming proofs. It checks if the keyset is known:
```python
            assert proof.id in self.keysets, (
                f"Keyset {proof.id} not known, can not verify DLEQ."
            )
```
If a user receives a token belonging to a deleted keyset, `verify_proofs_dleq` fails because the deleted keyset was omitted from `self.keysets`, aborting the operation.

callebtc (approved):

PR #801, opened 2025-10-19
feat: async melt requests

a1denvalu3:
## Summary
This PR adds support for `prefer_async` option in the `POST /v1/melt/bolt11` request body. If a request contains this option set to true, the Mint no longer blocks until the operation is completed, but rather it schedules a task for the backend and returns immediately with a `PENDING` melt quote response.

Relevant specification PR is [here](https://github.com/cashubtc/nuts/pull/285).

### Changes
- Added `prefer_async` field to `PostMeltRequest` model.
- Updated `Wallet.melt` to support `prefer_async` argument.
- Updated Mint's `POST /v1/melt/bolt11` endpoint to handle `prefer_async` from the request body.

### DIFF
--- a/cashu/mint/router.py
+++ b/cashu/mint/router.py
@@ -327,9 +327,11 @@ async def melt(request: Request, payload: PostMeltRequest) -> PostMeltQuoteRespo
     """
     Requests tokens to be destroyed and sent out via Lightning.
     """
-    logger.trace(f"> POST /v1/melt/bolt11: {payload}")
+    prefer = request.headers.get("Prefer")
+    logger.trace(f"> POST /v1/melt/bolt11: {payload}" + (f" with Prefer header set to {prefer}" if prefer else ""))
     resp = await ledger.melt(
-        proofs=payload.inputs, quote=payload.quote, outputs=payload.outputs
+        proofs=payload.inputs, quote=payload.quote, outputs=payload.outputs,
+        prefer_async=(True if prefer and "respond-async" in prefer.lower() else False)
     )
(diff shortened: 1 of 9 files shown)


a1denvalu3:
> found an issue (likely race condition) that causes a db exception when you call `get_melt_quote()`. we need to figure out why and should throw the error in the mint rather than in the database. will get back to this.

I have fixed this.

minibits-cash:
Sorry I am a little late to the discussion (if the NUT update would be still in the making I'd likely propose to have sync or async as a property of a melt method, to avoid all optionality here). Now with the custom header in place, I'd vote for having either sync or async behavior depending on header - and just handle internal bolt11 melts as async from outside POV if that's what wallet asked for.

The melt handling from wallet POV is already now full of traps and limit the number of possible behaviors would be a good choice.

cashu/mint/router.py
```diff
@@ -327,9 +327,13 @@ async def melt(request: Request, payload: PostMeltRequest) -> PostMeltQuoteRespo
     """
     Requests tokens to be destroyed and sent out via Lightning.
     """
-    logger.trace(f"> POST /v1/melt/bolt11: {payload}")
+    prefer_async = False
+    if payload.options and payload.options.prefer_async:
```
robwoodgate: Not sure using options is right? The new spec defines it at top level:

```json
{
  "quote": <str>,
  "inputs": <Array[Proof]>,
  "prefer_async": <bool> // optional: false if omitted
  // Additional method-specific fields may be required
}
```

PR #824, opened 2025-11-09
Add LN Markets Lightning backend integration

Forte11Cuba:
Adds LN Markets as a new Lightning Network backend, enabling mints to operate using LN Markets accounts with full Bitcoin Lightning capabilities.

Lightning Operations:
- Mint: Deposit sats via Lightning invoices
- Melt: Withdraw sats by paying Lightning invoices
- LNURL payments supported
- Lightning Address payments supported
- Real-time balance checking and payment status verification

Authentication:
- HMAC-SHA256 signature (LN Markets API v3)
- Requires API key, secret, and passphrase

Configuration
Add to .env:
MINT_BACKEND_BOLT11_SAT=LNMarketsWallet
MINT_LNMARKETS_KEY=your_api_key
MINT_LNMARKETS_SECRET=your_api_secret
MINT_LNMARKETS_PASSPHRASE=your_passphrase
MINT_LNMARKETS_ENDPOINT=https://api.lnmarkets.com/v3

Payments from this mint to another LN Markets account do not work due to LN Markets platform limitations that prevent circular payments within the same service. Payments to external Lightning nodes, LNURL, and Lightning addresses work correctly.

### DIFF
--- a/cashu/core/settings.py
+++ b/cashu/core/settings.py
@@ -103,9 +103,9 @@ class MintBackends(MintSettings):
     mint_strike_key: Optional[str] = Field(default=None)
     mint_blink_key: Optional[str] = Field(default=None)
 
-    mint_lnmarkets_key: str = Field(default=None)
-    mint_lnmarkets_secret: str = Field(default=None)
-    mint_lnmarkets_passphrase: str = Field(default=None)
+    mint_lnmarkets_key: Optional[str] = Field(default=None)
+    mint_lnmarkets_secret: Optional[str] = Field(default=None)
+    mint_lnmarkets_passphrase: Optional[str] = Field(default=None)
     mint_lnmarkets_endpoint: str = Field(
         default="https://api.lnmarkets.com/v3"
     )
--- a/cashu/lightning/lnmarkets.py
+++ b/cashu/lightning/lnmarkets.py
@@ -88,9 +88,17 @@ class LNMarketsWallet(LightningBackend):
         self.unit = unit
 
         self.endpoint = settings.mint_lnmarkets_endpoint
-        self.api_key = settings.mint_lnmarkets_key
-        self.api_secret = settings.mint_lnmarkets_secret
-        self.api_passphrase = settings.mint_lnmarkets_passphrase
+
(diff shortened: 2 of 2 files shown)


.env.example
```diff
@@ -112,6 +112,15 @@ MINT_BLINK_KEY=blink_abcdefgh
...
 MINT_STRIKE_KEY=ABC123
 
+# Use with LNMarkets for BTC
+MINT_LNMARKETS_KEY=your_lnmarkets_key
+MINT_LNMARKETS_SECRET=your_lnmarkets_secret
+MINT_LNMARKETS_PASSPHRASE=your_lnmarkets_passphrase
+# LN Markets API endpoint
+# - Production: https://api.lnmarkets.com/v3
+# - Testnet: https://api.testnet4.lnmarkets.com/v3
+MINT_LNMARKETS_ENDPOINT=https://api.lnmarkets.com/v3
```
callebtc: This shouldn't be in the .env.example IMO, but it's fine to leave it as a config param for development.

callebtc (commented):
Thank you! What's the state of this PR? Do fees work correctly? Would it be possible to add tests for the backend with mocked responses like we do for the blink backend? (see cashu/tests/mint/test_mint_lightning_blink.py)

The blink tests are not the best but they demonstrate how to use the mock library.

Any updates from LNMarkets if this could be used as a fiat mint too?

Forte11Cuba:
Hi, thank you very much for the comments I hadn’t seen this review.
I’ll start working on it and add tests similar to the ones in the Blink backend using mocked responses.

Regarding fiat support, it would be amazing, but LNMarkets hasn’t enabled the required endpoint yet. You can open a short or do a swap after receiving the sats, but I think this is problematic for mints because of volatility: the value of the tokens at the moment of the request could differ from the value at the moment the sats are actually received.

In API v2 there was an endpoint that allowed receiving sats and automatically converting them to sUSD while freezing the price for a few minutes, effectively absorbing volatility but this endpoint is not available in v3. I’ve contacted them about it, but it doesn’t seem to be a priority for them right now. Hopefully they reconsider.

PR #837, opened 2025-11-22
fix: add --version flag support to mint CLI

hsdredgun:
Previously, passing --version to the mint CLI would cause a TypeError
  because the flag was passed through to uvicorn.Config which doesn't
  accept a version parameter.

  This adds proper version option support using click's version_option
  decorator.

### DIFF
--- a/cashu/mint/main.py
+++ b/cashu/mint/main.py
@@ -4,7 +4,7 @@ import click
 import uvicorn
 from click import Context
 
-from ..core.settings import settings, VERSION
+from ..core.settings import VERSION, settings
 
 
 @click.command(


callebtc:
love these PRs thank you very much

cashu/mint/main.py
```diff
@@ -4,7 +4,7 @@
 import uvicorn
 from click import Context
 
-from ..core.settings import settings
+from ..core.settings import settings, VERSION
```
prusnak: I think this should fix the lint error in the CI  (unsorted imports)

```suggestion
from ..core.settings import VERSION, settings
```

a1denvalu3 (approved):

PR #839, opened 2025-11-25
Mint should check pending proofs and expired quotes regularly

### LINKED ISSUE #596: [Feature request] Mint should check pending proofs and expired quotes regularly
A task that runs once an hour or so to check all pending proofs, and expired mint and melt quotes, and deletes them / resolves them.

KvngMikey:
fixes #596

### Summary
- This PR implements proper expiration handling for mint and melt quotes that remain in the PENDING state beyond their valid lifetime. It introduces a background sweep that marks these stale quotes as EXPIRED, ensuring they can no longer be redeemed or incorrectly matched by later processing.

### Changes
- Pending mint/melt quotes were never expired, which could accumulate indefinitely and cause incorrect accounting or reuse, so now we expire them
- The system had no centralized regular task to sweep and update expired quotes, this was introduced to handle this.
- All test passs.

### Question
- Should we add an expiry column to mint_quotes and melt_quotes? Right now, expiry is determined implicitly using:
created_time + EXPIRY_SECONDS (5minutes, a fixed lifespan and global timeout)..
Please advise which approach you prefer so I can update accordingly.

### DIFF
--- a/cashu/mint/crud.py
+++ b/cashu/mint/crud.py
@@ -1053,4 +1053,4 @@ class LedgerCrudSqlite(LedgerCrud):
             """,
             {"checking_id": checking_id},
         )
-        return [MeltQuote.from_row(row) for row in results] # type: ignore
+        return [MeltQuote.from_row(row) for row in results]  # type: ignore


cashu/mint/crud.py
```diff
@@ -20,6 +23,7 @@
     Database,
 )
 
+EXPIRY_SECONDS = 300  #5 minutes
```
a1denvalu3: This shouldn't be hardcoded, but rather expsosed in the settings

cashu/mint/ledger.py
```diff
@@ -156,12 +158,28 @@ async def _startup_keysets(self) -> None:
...
-            await asyncio.sleep(settings.mint_regular_tasks_interval_seconds)
-        except Exception as e:
-            logger.error(f"Ledger regular task failed: {e}")
-            await asyncio.sleep(60)
+        """
+        Runs periodic ledger maintenance tasks forever.
+        This function intentionally loops forever and is designed to be scheduled as a Task.
+        """
+        logger.info("Starting ledger regular tasks loop")
+        while True:
```
a1denvalu3, line 164: Are you also fixing a bug here? I think this task was always supposed to loop forever.

cashu/mint/crud.py
KvngMikey: okay, thank you for this @lollerfirst...
can you help me with any pointers on how i should go about it? right now i just have it as 5mins...

that's why i asked this:
"Should we add an expiry column to mint_quotes and melt_quotes? Right now, expiry is determined implicitly using:
created_time + EXPIRY_SECONDS (5minutes, a fixed lifespan and global timeout)..
Please advise which approach you prefer so I can update accordingly." 

would appreciate your feedback...

cashu/mint/ledger.py
KvngMikey: yes, the old code here had a latent bug, the background sweep only ran once on initialisation rather than repeatedly, it ran once then slept after the sleep the function returned (completed), now it runs forever until cancelled, and I added a couple of safety improvements.

cashu/mint/crud.py
KvngMikey: @lollerfirst.
I've now moved the hard-coded constant value into the settings layer exposing the new field so operators can configure the expiry window, and all CRUD expiry logic now reads from settings.mint_quote_expiry_seconds.

Let me know if you'd prefer a different naming/location, but the setting is now fully configurable as requested, thank you.

KvngMikey:
based on a conversation had with @callebtc, I’ve updated the PR so that _run_regular_tasks() strictly runs as a simple infinite loop, performing the periodic check and then sleeping. All other logic has been removed from this function as requested to be implemented after a future conversation, thank you

cashu/mint/ledger.py
```diff
@@ -156,12 +158,27 @@ async def _startup_keysets(self) -> None:
...
```
callebtc, line 164: it would have been enough to simply add a `while True` here instead of all of the below

KvngMikey: done

a1denvalu3 (approved):

callebtc (approved):
thank you!

PR #841, opened 2025-11-28
feat: Add Spark wallet support with Breez Nodeless SDK

aljazceru:
- Adds SparkWallet backend
- Fixes docker-compose.yaml file to only default to fakewallet if the env variables are not present


cashu/lightning/spark.py
```diff
@@ -0,0 +1,524 @@
...
+        if BreezSdk is None:
+            raise Unsupported("breez-sdk-spark is required for SparkWallet")
+
+        self.assert_unit_supported(unit)
+        self.unit = unit
+
+        assert settings.mint_spark_api_key, "MINT_SPARK_API_KEY not set"
+        assert settings.mint_spark_mnemonic, "MINT_SPARK_MNEMONIC not set"
+
+        network_name = getattr(settings, "mint_spark_network", "mainnet").lower()
```
TheRealCheebs, line 188: You have these defaults already defined in settings.py:
```
    mint_spark_network: str = Field(default="mainnet")
    mint_spark_storage_dir: str = Field(default="data/spark")
    mint_spark_connection_timeout: int = Field(default=30)
    mint_spark_retry_attempts: int = Field(default=3)
```
We should manage them in one place.

aljazceru: ah yeah, good catch, i'll remove the duplicates

aljazceru: @TheRealCheebs I've finally looked into this properly, technically its following patterns from other backends that have settings define in settings.py but actually validate the needed credentials in their respective files

https://github.com/cashubtc/nutshell/blob/a59f369877c6565f16f700e3cbc5ff85e33e24c6/cashu/lightning/corelightningrest.py#L53

https://github.com/cashubtc/nutshell/blob/a59f369877c6565f16f700e3cbc5ff85e33e24c6/cashu/lightning/lndrest.py#L70

TheRealCheebs: I think those two are showing the pattern you have for the mint_spark_api_key and mint_spark_mnemonic; raising an exception or asserting when nothing is set. What I am calling out is you set defaults in settings.py for mint_spark_network, mint_spark_storage_dir,... , then you have defaults set again in the third parameter of getattr. The getattr defaults should be dropped as they are duplicates, and defaults should be managed in one place.

docker-compose.yaml
```diff
@@ -7,11 +6,13 @@ services:
...
+      - .env
     environment:
-      - MINT_BACKEND_BOLT11_SAT=FakeWallet
-      - MINT_LISTEN_HOST=0.0.0.0
-      - MINT_LISTEN_PORT=3338
-      - MINT_PRIVATE_KEY=TEST_PRIVATE_KEY
+      - MINT_BACKEND_BOLT11_SAT=${MINT_BACKEND_BOLT11_SAT:-FakeWallet}
+      - MINT_LISTEN_HOST=${MINT_LISTEN_HOST:-0.0.0.0}
+      - MINT_LISTEN_PORT=${MINT_LISTEN_PORT:-3338}
+      - MINT_PRIVATE_KEY=${MINT_PRIVATE_KEY:-TEST_PRIVATE_KEY}
```
a1denvalu3, line 15: unrelated changes?

Dockerfile
```diff
@@ -12,3 +12,4 @@ WORKDIR /app
 COPY . .
 RUN poetry config virtualenvs.create false
 RUN poetry install --no-dev --no-root
+
```
a1denvalu3: ?

docker-compose.yaml
aljazceru: yeah kinda, it makes the docker file actually work properly (vars taken from .env with sane defaults). otherwise you either need docker-compose.override.yaml or you always have a diff of the file in the repo. so its an improvemen, but it should possibly be in a separate PR :)

cashu/lightning/spark.py
```diff
@@ -0,0 +1,524 @@
...
+    )
+    from breez_sdk_spark import breez_sdk_spark as spark_bindings
+except ImportError as exc:  # pragma: no cover - optional dependency
+    logger.warning("Breez Spark SDK not available - Spark backend disabled: %s", exc)
+    BreezSdk = None  # type: ignore[assignment]
+    EventListener = object  # type: ignore[assignment]
+    PaymentMethod = None
+    PaymentType = None
+    SparkPaymentStatus = None
+    spark_bindings = None
```
a1denvalu3, line 56: We do not try and catch imports. We make sure they're added to the package.

PR #842, opened 2025-11-30
p2pk both spend paths possible after timelock

callebtc:
addresses https://github.com/cashubtc/nuts/pull/315

### DIFF
--- a/cashu/core/nuts/nut14.py
+++ b/cashu/core/nuts/nut14.py
@@ -18,20 +18,26 @@ def verify_htlc_spending_conditions(
     if not secret.kind or secret.kind != SecretKind.HTLC.value:
         raise TransactionError("not an HTLC secret.")
     htlc_secret = HTLCSecret.from_secret(secret)
-    # hash lock
-    if not proof.htlcpreimage:
-        raise TransactionError("no HTLC preimage provided")
-    # verify correct preimage (the hashlock) if the locktime hasn't passed
+    preimage = proof.htlcpreimage
     now = time.time()
-    if not htlc_secret.locktime or htlc_secret.locktime > now:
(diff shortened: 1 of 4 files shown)


cashu/core/nuts/nut14.py
```diff
@@ -18,20 +18,26 @@ def verify_htlc_spending_conditions(
...
+                htlc_secret.data
+            ):
                 raise TransactionError("HTLC preimage does not match.")
         except ValueError:
             raise TransactionError("invalid preimage for HTLC: not a hex string.")
-    return True
+        return True
+
+    # Refund path unlocks once the timelock has expired
+    if locktime_passed:
```
TheRealCheebs, line 40: nit on flow readability. Before the preimage checks were conditional on not being past the locktime, but this isn't the case now. So to me it reads weird to have the timelock code split around the preimage block.
```
    now = time.time()
    locktime = htlc_secret.locktime
    locktime_passed = locktime is not None and locktime <= now
    if locktime_passed:
```

tests/wallet/test_wallet_htlc.py
```diff
@@ -518,21 +547,22 @@ async def test_htlc_n_sigs_refund_locktime(wallet1: Wallet, wallet2: Wallet):
...
 
-    # # Wait for locktime to pass
-    # await asyncio.sleep(2)
+    # Should fail because we need 2 signatures
+    await assert_err(
+        wallet1.redeem(send_proofs_copy),
+        "HTLC preimage does not match.",
+    )
 
+    # NOW: try to redeem via locktime path but with only 1 signature
```
TheRealCheebs, line 561: double comment?

PR #843, opened 2025-12-03
refactor: move to coincurve

### LINKED ISSUE #592: migrate from secp256k1 to coincurve
i think we should just migrate to coincurve - it is well maintained (last secp256k1 release was in 2021) - and even CLN uses coincurve and not secp256k1 for their python plugins

a1denvalu3:
Migrate from `secp256k1-py` to `coincurve` bindings for the secp256k1 library.

fixes #592


cashu/core/crypto/secp.py
```diff
@@ -48,6 +46,6 @@ def to_data(self):
 PublicKey.__add__ = PublicKeyExt.__add__  # type: ignore
 PublicKey.__neg__ = PublicKeyExt.__neg__  # type: ignore
 PublicKey.__sub__ = PublicKeyExt.__sub__  # type: ignore
-PublicKey.mult = PublicKeyExt.mult  # type: ignore
+PublicKey.__mul__ = PublicKeyExt.__mul__  # type: ignore
```
TheRealCheebs, line 49: dumb comment: but it's nice when things fit nicely.  :smile:

TheRealCheebs:
This PR addresses https://github.com/cashubtc/nutshell/issues/592

cashu/core/nuts/nut20.py
```diff
@@ -35,7 +38,7 @@ def verify_mint_quote(
     public_key: str,
     signature: str,
 ) -> bool:
-    pubkey = PublicKey(bytes.fromhex(public_key), raw=True)
+    pubkey = PublicKeyXOnly(bytes.fromhex(public_key)[1:])
```
callebtc, line 41: Is there a PublicKeyCompressed in coincurve? Should be a tiny bit faster.

a1denvalu3: I followed https://ofek.dev/coincurve/api/ and it doesn't appear to have it.

TheRealCheebs:
@prusnak From a line of Gemini questioning. It was pointing to Python-to-C translation layer and object lifecycle management as the slowdown. And it was stating that bench-marking is "over-blowing" the slowness vs what anyone would see in  real world use cases. 

> 1. Object Instantiation vs. Static Calls
> 
> The primary difference is that secp256k1-py often allowed for "rawer" access to functions, whereas coincurve is strictly object-oriented.
> 
> The Hit: Every time you create a PrivateKey or PublicKey object in coincurve, Python has to allocate memory, run the __init__ method, and perform validation checks (like ensuring the key is within the valid curve range).
> 
> The Scale: If you are signing 10,000 messages and creating a new PrivateKey object for each one instead of reusing a single instance, the overhead of Python object creation will far outweigh the microseconds spent on the actual C-level ECDSA math.
> 
> 2. Defensive Validation (The "Robustness" Cost)
> coincurve is designed to be "safest by default."
> 
> Point Validation: When you load a public key, coincurve doesn't just treat it as a blob of bytes. It calls secp256k1_ec_pubkey_parse to ensure the point actually lies on the curve.
> 
> Normalization: It often performs extra checks to ensure signatures are in the "Lower-S" form (to prevent malleability), which is a requirement for modern blockchains like Ethereum but adds a few extra logic gates in the wrapper code.
> 
> Comparison: secp256k1-py frequently bypassed these checks if you used the lower-level functional interface, essentially "trusting" that your input data was already sanitized.
> 
> 3. CFFI vs. C-Extensions (The Bridge)
> Both libraries use CFFI (C Foreign Function Interface), but they use it differently:
> 
> Type Marshalling: coincurve does more "translation" between Python types (like bytes) and C types (like unsigned char*). It ensures that memory is correctly pinned and that pointers are handled safely to avoid segfaults.
> 
> The Context Object: libsecp256k1 requires a "context" (a pre-computed table of values) to run fast. coincurve manages a global context for you. While this is generally a performance win, the internal logic to ensure that context is thread-safe and properly initialized adds a tiny layer of "bookkeeping" to every call.



And the areas to clean up if we see real world performance hits would be around

> Reuse Objects: Never call PrivateKey(seed) inside a loop. Create the object once and call .sign() repeatedly.
> 
> Batch Verification: If you are verifying many signatures, look for batching opportunities. coincurve is optimized for repeated use of the same PublicKey object.
> 
> Avoid Re-parsing: If you have public keys in byte format, parse them into PublicKey objects once and store the objects in a cache rather than re-parsing the bytes every time you need to verify a signature.

It looks like Coincurve is the definitive standard now for python bindings to the secp256k1 curve, and I think the trade offs between safety and robustness/maintained code between the two should trump the performance hits we see when thrown in sandboxed benchmarks.

callebtc:
LGTM

PR #872, opened 2026-01-17
Remove Redis cache from swap endpoint

### LINKED ISSUE #847: Redis cache prevents swap outputs from being stored
Restore wallet tests fail because swap outputs are missing. Tests expect 96/128/84 sat but only get 64 sat (initial mint).

Failing Tests

- `test_restore_wallet_after_swap_to_send` - Expected: 96 sat, Got: 64 sat
- `test_restore_wallet_after_send_and_receive` - Expected: 96 sat, Got: 64 sat
- `test_restore_wallet_after_send_and_self_receive` - Expected: 128 sat, Got: 64 sat
- `test_restore_wallet_after_send_and_self_receive_nonquadratic_value` - Expected: 84 sat, Got: 64 sat

What's happening
1. Mint 64 sat → creates 15 promises (works)
2. Swap 32 sat → should create additional change output promises
3. Reset wallet DB → clears local wallet database
4. Restore → only finds 15 promises (64 sat), missing the swap change outputs

fresh3nough:
The Redis cache decorator was preventing swap outputs from being stored in the database when the same request was made multiple times. This caused wallet restoration tests to fail because the promises weren't found in the database.

Swap operations should never be cached since each swap is unique and must be processed and stored in the database, even if the inputs and outputs are identical.

Fixes #847

### DIFF
--- a/cashu/mint/router.py
+++ b/cashu/mint/router.py
@@ -349,7 +349,6 @@ async def melt(request: Request, payload: PostMeltRequest) -> PostMeltQuoteRespo
     ),
 )
 @limiter.limit(f"{settings.mint_transaction_rate_limit_per_minute}/minute")
-@redis.cache()
 async def swap(
     request: Request,
     payload: PostSwapRequest,


a1denvalu3:
> The Redis cache decorator was preventing swap outputs from being stored in the database when the same request was made multiple times. This caused wallet restoration tests to fail because the promises weren't found in the database.

NAK. The request is cached only if it's successful. If it's not successful, it is not cached. If the request is successful, a second instance of the same request should be IDEMPOTENT (NUT-19), which means it should yield the exact same response. This is done because a wallet might issue the exact same request multiple times due to network errors.

fresh3nough:
The key issue: The decorator caches ANY response that doesn't raise an exception. It doesn't distinguish between successful and failed operations. More importantly, even for successful operations, it caches the response at the HTTP layer which can happen before database transactions commit.

Current solution (which is correct):

We've already removed @redis.cache() from the swap endpoint because:
•  Database-level idempotency handles duplicate requests
•   No risk of caching responses before DB commits
•   The restore endpoint always finds promises in the database
•   Simpler and more reliable than HTTP-level cachin

a1denvalu3:
https://github.com/cashubtc/nutshell/issues/847#issuecomment-3777598466

fresh3nough:
@a1denvalu3 @KvngMikey requesting a review whenever one of you get a chance

a1denvalu3:
NACK. I've already discussed the solution isn't removing the cache from the swap endpoint. See https://github.com/cashubtc/nutshell/issues/847#issuecomment-3777598466

PR #876, opened 2026-01-22
fix: redis startup connection check

a1denvalu3:
- Moves the Redis connection check from the background task in `__init__` to the application startup `lifespan`.
- Ensures the application verifies the Redis connection before accepting requests when `MINT_REDIS_CACHE_ENABLED` is true.


KvngMikey (approved):

callebtc (approved):
Thank you, good fix

PR #878, opened 2026-01-26
feat: batch minting

a1denvalu3:
## Summary
Implemented NUT-XX Batch Minting specification.

This PR adds support for:
- `POST /v1/mint/quote/{method}/check`: Batch checking of quote states.
- `POST /v1/mint/{method}/batch`: Atomic batch minting of multiple quotes.
- NUT-20 signature verification for batched quotes.

The implementation ensures that all quotes in a batch are processed atomically. If any verification fails or any quote cannot be minted, the entire batch is rejected.

### DIFF
--- a/cashu/mint/db/write.py
+++ b/cashu/mint/db/write.py
@@ -223,9 +223,16 @@ class DbWriteHelper:
         if not quote_ids:
             return []
 
-        # Lock the entire table for now as row-level locking with IN clauses requires db changes
+        # Use dynamically generated IN clause for row-level locking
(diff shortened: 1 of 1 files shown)


cashu/core/nuts/nuts.py
```diff
@@ -17,3 +17,4 @@
 CLEAR_AUTH_NUT = 21
 BLIND_AUTH_NUT = 22
 METHOD_BOLT11_NUT = 23
+BATCH_NUT = 333
```
KvngMikey: i'm guessing this '333' is a placeholder as the nut says 'xx'... can we document that this is a placeholder so i remember to come back and update it ?

PR #880, opened 2026-01-28
fix: set/unset melt quote and associated proofs in same transaction

a1denvalu3:
This PR adds a test case that reproduces a race condition in the melt operation. When two concurrent melt requests attempt to spend the same proofs, one fails as expected, but leaves the proofs in a PENDING state indefinitely.

### DIFF
--- a/tests/mint/test_mint_auth_server_unit.py
+++ b/tests/mint/test_mint_auth_server_unit.py
@@ -196,12 +196,11 @@ async def test_verify_blind_auth_invalidates_on_success_and_unsets_pending():
         async def _unset_proofs_pending(self, proofs, keysets):
             calls["unset"] = proofs[0].secret
 
-    async def invalidate_proofs(*, proofs):
-        calls["invalidated"] = proofs[0].secret
+        async def invalidate_proofs(self, *, proofs, keysets):
+            calls["invalidated"] = proofs[0].secret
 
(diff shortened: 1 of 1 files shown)


cashu/mint/db/write.py
```diff
@@ -322,3 +336,122 @@ async def _store_melt_quote(self, quote: MeltQuote):
...
+        Args:
+            quote (MeltQuote): Melt quote to set as PAID.
+            proofs (List[Proof]): Proofs to invalidate (spend).
+            keysets (Dict[str, MintKeyset]): Keysets for updating balances/fees.
+            keyset_fees (Dict[str, int]): Fees paid per keyset.
+        """
+        quote_copy = quote.model_copy()
+
+        async with self.db.get_connection(
+            lock_table="proofs_pending",
```
TheRealCheebs, line 480: do we want to add a lock_timeout to these?  The calls we are wrapping did have 1 set prior, but now that we pass the connection through its using the default of 5.

cashu/mint/ledger.py
```diff
@@ -863,6 +863,13 @@ async def melt(
                     f"output unit {outputs_unit.name} does not match quote unit {melt_quote.unit}"
                 )
 
+        # make sure that the proofs are in the same unit as the quote
```
KvngMikey: love this.

Issue #883, opened 2026-01-30
Implement new protected endpoint matching rules


robwoodgate:
https://github.com/cashubtc/nuts/pull/334

PR #885, opened 2026-02-04
feat: protected endpoint wildcard matching rules

robwoodgate:
implements https://github.com/cashubtc/nuts/pull/334
closes: https://github.com/cashubtc/nutshell/issues/883

### DIFF
--- a/cashu/core/mint_info.py
+++ b/cashu/core/mint_info.py
@@ -1,5 +1,4 @@
 import json
-import re
 from typing import Any, Dict, List, Optional
 
 from pydantic import BaseModel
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -55,6 +55,7 @@ ruff = "^0.7.1"
 mypy = "^1.13.0"
 pre-commit = "^4.0.1"
 types-redis = "^4.6.0"
+greenlet = "^3.3.1"
```
a1denvalu3: Is this necessary?

robwoodgate: All the tests failed without it when I span up the environment. I assume it was previously a sub-dep of something else, and no longer installed on a clean install... but the tests relied on it.

Try removing your pypoetry cache and start from scratch. Every test fails with something like:

```python
ERROR tests/wallet/test_wallet_p2pk.py::test_p2pk_locktime_with_3_of_3_refund_pubkeys - ValueError: the greenlet library is required to use this function. No modul...
ERROR tests/wallet/test_wallet_p2pk_methods.py::test_create_p2pk_lock_default - ValueError: the greenlet library is required to use this function. No modul...
ERROR tests/wallet/test_wallet_p2pk_methods.py::test_create_p2pk_lock_with_options - ValueError: the greenlet library is required to use this function. No modul...
```

callebtc: what is this about? we shouldn't add libraries if we don't know what they're doing.

robwoodgate: The test errors are pretty specific that greenlet is a dependency, but it was not in the project:


```
greenlet is a lightweight coroutine/context-switching library for Python.

In practice, it’s often used by frameworks (especially SQLAlchemy async internals) to “jump” between sync-style code and async event-loop code without full threads.

So that error usually means:

• some code path expects SQLAlchemy async/sync bridging that relies on greenlet
• but greenlet isn’t installed in that environment
```

But it's not core to the actual change, I added it to ensure tests ran properly.

robwoodgate: Marking as resolved as you reverted that incidental change

PR #897, opened 2026-02-19
chore: verify version consistency in build workflows

a1denvalu3:
## Summary
This PR adds a verification step to the Docker and PyPI build workflows to ensure that the release tag version matches the versions defined in the codebase.

If any version mismatch is detected, the workflow will fail early, preventing inconsistent releases.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ authors = ["calle <callebtc@protonmail.com>"]
 license = "MIT"
 
 [tool.poetry.dependencies]
-python = "^3.11"
+python = "^3.10"
 SQLAlchemy = {version = "^2.0.35", extras = ["asyncio"]}
 click = "^8.1.7"
 pydantic = "^2.5.3"


scripts/check_version.py
```diff
@@ -0,0 +1,88 @@
...
+        print(f"File not found: {pyproject_path}")
+        sys.exit(1)
+
+    # limiting search to [tool.poetry] section is harder with regex, 
+    # but usually version is early in the file.
+    # We look for 'version = "..."'
+    match = re.search(r'^version = "(.*?)"', content, re.MULTILINE)
+    if match:
+        return match.group(1)
+    raise ValueError("Could not find version in pyproject.toml")
```
prusnak: We can use `tomllib` since Python 3.11:


``` python
import tomllib

with open("pyproject.toml", "rb") as f:
    data = tomllib.load(f)

version = data["tool"]["poetry"]["version"]
```

but I agree that expecting first `version = ` to be the correct one is reasonable assumption and probably also future-proof when we switch from poetry to uv

a1denvalu3: nice.

PR #900, opened 2026-02-21
Remove deprecated promises field from PostRestoreResponse

### LINKED ISSUE #898: Remove the old promises field from PostRestoreResponse
The promises field was deprecated in 0.15.1. It was kept for compatibility with old cashu-ts based clients removed in [v2.3.0](https://github.com/cashubtc/cashu-ts/releases/tag/v2.3.0). Its been more than a year so it's time to remove this.

CDK is also cleaning this up [here](https://github.com/cashubtc/cdk/issues/1644)

coldbrewnosugar:
## Summary
- Remove the deprecated `promises` field from `PostRestoreResponse` and its `__init__` override that duplicated `signatures` into `promises`
- Remove the backwards compatibility shim in the wallet API that copied `promises` back into `signatures`
- Update the deprecated API restore test to assert on `signatures` instead of `promises`

The `promises` field was deprecated in v0.15.1 for compatibility with old cashu-ts clients. cashu-ts completed the same rename in [v2.3.0](https://github.com/cashubtc/cashu-ts/releases/tag/v2.3.0) (PR [#234](https://github.com/cashubtc/cashu-ts/pull/234)) over a year ago. CDK is also cleaning this up in [#1644](https://github.com/cashubtc/cdk/issues/1644).

Closes #898

### DIFF
--- a/tests/mint/test_mint_api_deprecated.py
+++ b/tests/mint/test_mint_api_deprecated.py
@@ -7,11 +7,8 @@ from cashu.core.models import (
     CheckSpendableRequest_deprecated,
     CheckSpendableResponse_deprecated,
     GetMintResponse_deprecated,
-    PostRestoreRequest,
-    PostRestoreResponse,
 )
 from cashu.mint.ledger import Ledger
-from cashu.wallet.crud import bump_secret_derivation
 from cashu.wallet.wallet import Wallet
 from tests.helpers import get_real_invoice, is_fake, is_regtest, pay_if_regtest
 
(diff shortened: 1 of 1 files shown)


tests/mint/test_mint_api_deprecated.py
```diff
@@ -348,12 +348,12 @@ async def test_api_restore(ledger: Ledger, wallet: Wallet):
         json=payload.model_dump(),
     )
     data = response.json()
-    assert "promises" in data
+    assert "signatures" in data
```
TheRealCheebs: This `test_api_restore` test should just be removed completely from here. The PostRestoreRequest is already tested in the `test_mint_api.py` with the `signatures` field. This test was to ensure the backwards compatibility was still valid in its deprecated state.

PR #901, opened 2026-02-22
fix restore_promises derivation-path mismatch on subset restores

b-l-u-e:
This PR fixes a wallet restore issue where restored proofs could receive the wrong `derivation_path` when the mint returns only a subset of requested outputs (or returns them in a non-zero index range/order).  
It also adds a regression test to ensure this mismatch cannot reoccur.

so here in `restore_promises`:

- `secrets` and `rs` were filtered using `matching_indices`
- but `derivation_paths` was not filtered before calling `_construct_proofs(...)`

#### 1) Wallet fix
In `cashu/wallet/wallet.py`:
- filtered `derivation_paths` using the same `matching_indices` used for `secrets`/`rs` before `_construct_proofs(...)`.
```diff
     secrets = [secrets[i] for i in matching_indices]
         rs = [rs[i] for i in matching_indices]
+        derivation_paths = [derivation_paths[i] for i in matching_indices]
         logger.debug(
             f"Restored {len(restored_promises)} promises. Constructing proofs."
         )
 ```

This restores 1:1 alignment among:
- `restored_promises`
- filtered `secrets`
- filtered `rs`
- filtered `derivation_paths`

#### 2) Regression test
In `tests/wallet/test_wallet_restore.py`:
- added `test_restore_promises_derivation_paths_subset`
- test verifies each restored proof’s `derivation_path` matches the path expected for that proof’s `secret`
- catches subset/non-zero-index restore mismatch scenarios.

### Test Plan
- `poetry run pytest tests/wallet/test_wallet_restore.py::test_restore_promises_derivation_paths_subset -v`

Note: While preparing this PR I hit a local pre-commit mypy error:
`tests/conftest.py:119: Argument "seed" to "Ledger" has incompatible type "str | None"; expected "str"`

not sure of the cause but maybe comes from here or rather maybe i am missing somethin???

```
git add tests/wallet/test_wallet_restore.py
git commit -s -m "test(wallet): add regression test for restore derivation-path mismatch"
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...........................................(no files to check)Skipped
check for added large files..............................................Passed
check docstring is first.................................................Passed
check json...........................................(no files to check)Skipped
debug statements (python)................................................Passed
mixed line ending........................................................Passed
check for case conflicts.................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

tests/conftest.py:119: error: Argument "seed" to "Ledger" has incompatible type "str | None"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

so i had to add an explicit guard in tests/conftest.py before constructing Ledger:
```diff
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -114,6 +114,7 @@ async def ledger():
             Unit.usd: lightning_backend_usd,
         },
     }
+    assert settings.mint_private_key is not None
     ledger = Ledger(
         db=Database("mint", settings.mint_database),
         seed=settings.mint_private_key,
```
also i struggled with ruff-format for the cashu/wallet/wallet.py and ended up making changes in other part of the code file which is unrelated to the issue solved

### DIFF
--- a/tests/wallet/test_wallet_restore.py
+++ b/tests/wallet/test_wallet_restore.py
@@ -394,3 +394,37 @@ async def test_restore_wallet_after_send_and_self_receive_nonquadratic_value(
     assert wallet3.balance == 108
     await wallet3.invalidate(wallet3.proofs, check_spendable=True)
     assert wallet3.balance == 64
+
+@pytest.mark.asyncio
+async def test_restore_promises_derivation_paths_subset(wallet3: Wallet):
+    """Restored proofs get derivation_path that matches their secret (subset restore)."""
+    await reset_wallet_db(wallet3)
+
+    # Advance counter so next mint uses indices 21+; restore then returns a subset.
+    await wallet3.restore_promises_from_to(wallet3.keyset_id, 0, 20)
+
+    mint_quote = await wallet3.request_mint(21)
+    await pay_if_regtest(mint_quote.request)
+    await wallet3.mint(21, quote_id=mint_quote.quote)
+    assert wallet3.balance > 0
+
+    restore_from, restore_to = 0, 40
+    secrets, _, derivation_paths = await wallet3.generate_secrets_from_to(
+        restore_from, restore_to
+    )
+    secret_to_path = dict(zip(secrets, derivation_paths))
+
+    await reset_wallet_db(wallet3)
+    await wallet3.load_proofs()
+    wallet3.proofs = []
+    assert wallet3.balance == 0
+    await wallet3.restore_promises_from_to(wallet3.keyset_id, restore_from, restore_to)
+
+    assert len(wallet3.proofs) >= 1
+    for proof in wallet3.proofs:
+        expected_path = secret_to_path.get(proof.secret)
+        assert expected_path is not None
+        assert proof.derivation_path == expected_path, (
+            f"derivation_path {proof.derivation_path!r} does not match secret "
+            f"(expected {expected_path!r})"
+        )


tests/conftest.py
```diff
@@ -114,6 +114,7 @@ async def start_mint_init(ledger: Ledger) -> Ledger:
             Unit.usd: lightning_backend_usd,
         },
     }
+    assert settings.mint_private_key is not None
```
a1denvalu3, line 117: why is this necessary?

callebtc: cos it could be None

tests/wallet/test_wallet_regtest_mpp.py
```diff
@@ -230,6 +230,10 @@ def mint_pay_mpp(invoice: str, amount: int, proofs: List[Proof]):
 async def test_regtest_pay_mpp_cancel_payment_pay_partial_invoice(
     wallet: Wallet, ledger: Ledger
 ):
+    # make sure that mpp is supported by the bolt11-sat backend
+    if not ledger.backends[Method["bolt11"]][wallet.unit].supports_mpp:
+        pytest.skip("backend does not support mpp")
+
```
a1denvalu3, line 236: This fix is not relevant to this PR. Will be merged with #948

b-l-u-e: its resolved thanks

callebtc (approved):

PR #913, opened 2026-03-05
fix: add greenlet as an explicit dependency

### LINKED ISSUE #899: Tests fail in newly set up environments
Tests fail when setting up a clean environment on Mac. The error is always something like:

```python
ERROR tests/wallet/test_wallet_p2pk.py::test_p2pk_locktime_with_3_of_3_refund_pubkeys - ValueError: the greenlet library is required to use this function. No modul...
ERROR tests/wallet/test_wallet_p2pk_methods.py::test_create_p2pk_lock_default - ValueError: the greenlet library is required to use this function. No modul...
ERROR tests/wallet/test_wallet_p2pk_methods.py::test_create_p2pk_lock_with_options - ValueError: the greenlet library is required to use this function. No modul...
```
Essentually, Greenlet is not installed as a dev dependency. I assume greenlet was previously a sub-dep of something else, which is why pre-existing environments do not see the issue.

You can reproduce by removing your pypoetry cache and starting install from scratch.

AI says: 

```
greenlet is a lightweight coroutine/context-switching library for Python.

In practice, it’s often used by frameworks (especially SQLAlchemy async internals) to “jump” between sync-style code and async event-loop code without full threads.

So that error usually means:

• some code path expects SQLAlchemy async/sync bridging that relies on greenlet
• but greenlet isn’t installed in that environment
```

wqxoxo:
Fixes #899.


pyproject.toml
```diff
@@ -8,6 +8,7 @@ license = "MIT"
 [tool.poetry.dependencies]
 python = "^3.10"
 SQLAlchemy = {version = "^2.0.35", extras = ["asyncio"]}
+greenlet = ">=1.0.0"
```
b-l-u-e: `greenlet = ">=1.0.0"` will install, but it’s a bit broad....perhaps  `>=3.0.0` or `^3.1.0`

wqxoxo: Good call, updated to `>=3.0.0`.

PR #920, opened 2026-03-21
fix: support both padded and unpadded base64url in AuthProof (NUT-22)

### LINKED ISSUE #919: Confirm NUT-22 authA serialization
https://github.com/cashubtc/nuts/pull/353 tightens up the specification, and requires mints to support both padded and unpadded forms of the base64_url `AuthProof` header. 

Please can we confirm Nutshell is compliant.

KvngMikey:
Fixes #919

The `AuthProof` serialization was using standard `Base64` instead of `base64url` as required by `NUT-22`. 
Additionally, per recently updated cashubtc/nuts#353, mints must accept both `padded` and `unpadded` forms of the `base64url AuthProof header.`

### Changes
- `to_base64`: use `urlsafe_b64encode` and strip trailing `=` padding to emit the compact form
- `from_base64`: use `urlsafe_b64decode` and re-pad before decoding so the mint accepts both padded and unpadded authA tokens.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,6 +1,6 @@
 import pytest
 
-from cashu.core.base import TokenV3, TokenV4, Unit
+from cashu.core.base import AuthProof, TokenV3, TokenV4, Unit
 from cashu.core.helpers import calculate_number_of_blank_outputs
 from cashu.core.secret import Secret, SecretKind, Tags
 from cashu.core.split import amount_split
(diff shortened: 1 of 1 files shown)


cashu/core/base.py
```diff
@@ -1428,7 +1428,9 @@ def from_base64(cls, base64_str: str):
             f"Token prefix not valid. Expected {cls.prefix}."
         )
         base64_str = base64_str[len(cls.prefix) :]
-        return cls.model_validate(json.loads(base64.b64decode(base64_str).decode()))
+        # Re-add padding if stripped, as urlsafe_b64decode requires it
+        padded = base64_str + "=" * (-len(base64_str) % 4)
```
a1denvalu3, line 1432: What's the logic for using `-len(base64_str)` ?

KvngMikey: this just computes how many `=` padding characters are needed to bring the string length up to the next multiple of 4, which is what base64url decoding requires, i think you did the same somewhere in the Nut18 file.

PR #958, opened 2026-03-30
feat(mint): implement NUT-29 batched minting

a1denvalu3:
## Summary
Implements the NUT-29 batched minting specification on the mint side.

### Details
- **Models**: Added `PostMintQuoteCheckRequest`, `PostMintBatchRequest`, and `PostMintBatchResponse`.
- **API Endpoints**: 
  - `POST /v1/mint/quote/bolt11/check`
  - `POST /v1/mint/bolt11/batch`
- **Ledger Logic**: Implemented `mint_quote_check` and `mint_batch` with strict validations for payment methods, units, states, amount balances, and NUT-20 aggregated signatures.
- **Mint Info**: Advertises `BATCH_MINT_NUT` (29) support with `max_batch_size` based on mint configurations.
- **Tests**: Added full test coverage for the new endpoints and validation logic in `tests/mint/test_mint_api_batch.py`.

*References: [NUT-29 Specification](https://github.com/cashubtc/nuts/blob/main/29.md)*


a1denvalu3:
Vulnerability: `mint_quote_check` fails to check Lightning backend for payment status

The newly introduced `mint_quote_check` endpoint (for batch checking mint quotes) fails to call `ledger.check_mint_quote()` to query the underlying Lightning node for the payment status.

In Cashu, mints typically do not run a background loop to constantly check the status of pending Lightning invoices. Instead, they rely on the client polling the quote endpoint. The original single-quote endpoint (`GET /v1/mint/quote/bolt11/{quote}`) correctly calls `ledger.check_mint_quote(mint_quote)` which reaches out to the Lightning node to verify the invoice and update the quote state in the database to `PAID`. 

However, the new batch `mint_quote_check` endpoint simply reads the quote from the local database:
```python
    async def mint_quote_check(self, payload: PostMintQuoteCheckRequest) -> List[MintQuote]:
        quotes: List[MintQuote] = []
        for quote_id in payload.quotes:
            quote = await self.get_mint_quote(quote_id) # Reads from DB only
            ...
            quotes.append(quote)
        return quotes
```
If a user pays the Lightning invoices and their wallet client exclusively polls the new `mint_quote_check` endpoint, the mint will never query the Lightning node. The quotes will remain stuck in the `UNPAID` state forever in the database, preventing the user from minting their tokens. The lightning payment will eventually settle and the user will lose their funds without receiving the tokens.

Exploit PoC:
```python
import asyncio
import httpx

async def poc():
    # 1. Request a mint quote
    quote_res = httpx.post("http://localhost:3337/v1/mint/quote/bolt11", json={"amount": 100, "unit": "sat"})
    quote_id = quote_res.json()["quote"]
    
    # 2. Pay the invoice (out of band)
    # The client pays the lightning invoice using their Lightning node...
    # (Simulate payment using fake wallet auto-settle or manual pay)
    
    # 3. Poll using the new batch endpoint
    check_res = httpx.post("http://localhost:3337/v1/mint/quote/bolt11/check", json={"quotes": [quote_id]})
    state = check_res.json()[0]["state"]
    
    # state will incorrectly remain "UNPAID" because check_mint_quote was never called!
    assert state == "UNPAID", "Vulnerability is present: Quote remains UNPAID"
    print("PoC successful: Quote stuck in UNPAID state.")
```

PR #981, opened 2026-04-18
fix: update default mint settings

a1denvalu3:
## Summary
Updates the default value of `mint_rate_limit_proxy_trust` to `False`.


cashu/core/settings.py
```diff
@@ -111,7 +111,7 @@ class MintLimits(MintSettings):
         description="IP-based rate limiter.",
     )
     mint_rate_limit_proxy_trust: bool = Field(
-        default=False,
+        default=True,
```
b-l-u-e, line 114: Commit 1 and commit 3 appear to apply the same changes  could you squash or remove the duplicate so the history is easier to follow?

PR #991, opened 2026-04-27
fix(mint): preserve change promises order in paid melt quotes

### LINKED ISSUE #986: get_blind_signatures_melt_id returns non-deterministic order
## get_blind_signatures_melt_id returns non-deterministic order

When a wallet checks a completed melt quote via GET /v1/melt/quote/bolt11/{id}, the change field contains blind signatures loaded from the promises table without an ORDER BY clause.

Also, it seems like maybe `get_blinded_messages_melt_id` in crud.py fetches the unsigned blinded messages from DB when reconstructing a quote response. These B_ values are then passed to `_generate_change_promises` which sorts amounts descending and assigns them to the DB-loaded messages. If they come back out of order, the wrong denominations get assigned to the wrong B_ values, producing mismatched signatures at the source. This means even the response from `get_melt_quote` can have wrong pairings.

The overall issue is that when reading from the db the ordering of outputs is not guaranteed and wallets expect the outputs to be in the same order as the blinded messages sent to the mint.

Related CDK issue: https://github.com/cashubtc/cdk/issues/1933

a1denvalu3:
Fixes #986 

## Summary
- Fixes a bug where `get_blinded_messages_melt_id` retrieved `outputs` from the database without an `ORDER BY` clause, leading to incorrectly mapped change signatures.
- Adds an `order_index` column to the `promises` table via a database schema migration (`m034_promises_order_index`).
- Updates `Ledger._store_blinded_messages` to record the original insertion order of the `outputs` correctly.
- Ensures `crud.get_blinded_messages_melt_id` and `crud.get_blind_signatures_melt_id` retrieve data sequentially via `ORDER BY order_index ASC`.
- Adds comprehensive unit and integration tests to verify the stable mapping and correct unblinding of multi-change denominations during delayed/asynchronous melt quote settlements.

### DIFF
--- a/cashu/mint/crud.py
+++ b/cashu/mint/crud.py
@@ -162,6 +162,7 @@ class LedgerCrud(ABC):
         mint_id: Optional[str] = None,
         melt_id: Optional[str] = None,
         swap_id: Optional[str] = None,
+        order_index: int = 0,
         conn: Optional[Connection] = None,
     ) -> None: ...
 
@@ -345,13 +346,14 @@ class LedgerCrudSqlite(LedgerCrud):
         mint_id: Optional[str] = None,
         melt_id: Optional[str] = None,
         swap_id: Optional[str] = None,
+        order_index: int = 0,
         conn: Optional[Connection] = None,
     ) -> None:
         await (conn or db).execute(
             f"""
             INSERT INTO {db.table_with_schema('promises')}
-            (amount, b_, id, created, mint_quote, melt_quote, swap_id)
-            VALUES (:amount, :b_, :id, :created, :mint_quote, :melt_quote, :swap_id)
+            (amount, b_, id, created, mint_quote, melt_quote, swap_id, order_index)
+            VALUES (:amount, :b_, :id, :created, :mint_quote, :melt_quote, :swap_id, :order_index)
             """,
             {
                 "amount": amount,
@@ -361,6 +363,7 @@ class LedgerCrudSqlite(LedgerCrud):
                 "mint_quote": mint_id,
                 "melt_quote": melt_id,
                 "swap_id": swap_id,
+                "order_index": order_index,
             },
         )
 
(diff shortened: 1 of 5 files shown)


cashu/mint/migrations.py
```diff
@@ -1234,3 +1234,13 @@ async def m033_add_issued_time_to_mint_quote(db: Database):
...
             f"ALTER TABLE {db.table_with_schema('mint_quotes')} ADD COLUMN issued_time TIMESTAMP"
         )
+
+async def m034_promises_order_index(db: Database):
+    """
+    Add order_index to promises table to preserve the original order of outputs.
+    """
+    async with db.connect() as conn:
+        await conn.execute(
+            f"ALTER TABLE {db.table_with_schema('promises')} ADD COLUMN order_index INTEGER DEFAULT 0"
```
KvngMikey, line 1244: here, all existing rows get `order_index = 0`,
If there's a pending melt_quote in the DB at upgrade time with multiple outputs, they all get 0 and the ordering is still broken for that quote. 

This is a narrow edge case but worth a comment in the migration at minimum.

cashu/mint/crud.py
```diff
@@ -345,13 +346,14 @@ async def store_blinded_message(
         mint_id: Optional[str] = None,
         melt_id: Optional[str] = None,
         swap_id: Optional[str] = None,
+        order_index: int = 0,
```
KvngMikey, line 349: if multiple call sites don't pass `order_index`, it will silently store `0` for all rows, then we have multiple rows with the same `order_index`.

I suggest the default be removed to make it required or we make default None 
`order_index: Optional[int] = None,`
with a DB Null fallback

a1denvalu3: I think having multiple `order_index` at 0 vs None doesn't change anything.

callebtc (approved):
ACK

KvngMikey (approved):

PR #997, opened 2026-05-03
fix: add ORDER BY id to blind signatures and blinded messages melt queries

### LINKED ISSUE #986: get_blind_signatures_melt_id returns non-deterministic order

Codepocketdev:
Fixes #986 - both SQL queries fetched rows without ORDER BY clause returning non-deterministic order. Wrong denominations were being assigned to wrong B_ values producing mismatched signatures on melt quote responses.

### DIFF
--- a/cashu/mint/crud.py
+++ b/cashu/mint/crud.py
@@ -375,6 +375,7 @@ class LedgerCrudSqlite(LedgerCrud):
             f"""
             SELECT * from {db.table_with_schema('promises')}
             WHERE melt_quote = :melt_id AND c_ IS NULL
+            ORDER BY id
             """,
             {"melt_id": melt_id},
         )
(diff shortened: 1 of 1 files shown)


a1denvalu3:
This is already fixed in #991

Codepocketdev:
Thanks for the heads up — didn't see #991 before opening this. Closing.

PR #1008, opened 2026-05-18
fix a variety of issues with SIG_ALL and its interaction with P2PK and HTLC

SatsAndSports:
Fixes Issue: #1009 which reports a variety of issues related to P2PK and HTLC and SIG_ALL. Even if some of those features have been partially implemented and tested together, problems arise when they are combined; for example, a post-locktime HTLC with SIG_ALL

_Starting with the mint:_

The existing code had NUT-10 code (i.e. P2PK and HTLC) in multiple places. In this PR, following the CDK architecture, all the NUT-10 logic is moved into a single boolean function. This unifies checking for swap and melt and minimizes code duplication.

The main functions in `cashu/mint/conditions.py` now are:
 - `_verify_transaction` is a new entry point for all verification of any swap or melt.
 - `_verify_input_output_spending_conditions` is an existing function that has been rewritten to verify the inputs and outputs (and quote, for a melt), looking just at the NUT-10 spending conditions (i.e. P2PK and HTLC). It returns True or raises an error. It delegates to various new helpers to deal with all the various P2PK or HTLC or SIG_ALL.
 -  `_verify_p2pk_or_htlc_spending_requirements` is the low level function to count valid signatures and, if necessary, the pre-image. This one function is reused across all the various contexts, for any NUT-10 secret that appears in the inputs of any swap or melt.

_DataStructures:_ This PR introduces a similar `SpendingRequirements` class as used in the CDK, to unify the requirements encoded in a P2PK/HTLC Secret.

_Wallet:_ The only wallet change here is to keep up with the new SIG_ALL message format.


SatsAndSports:
~There's still a lot of AI slop here. My first priority is completing all the tests and ensuring they are good and correct; the CDK tests are very helpful here as the CDK has a broad range of tests for this, then I'll review and tidy up the implementation here~

Update a few days later: ready for review now

cashu/core/p2pk.py
```diff
@@ -1,13 +1,16 @@
...
-from typing import Union
+from typing import TYPE_CHECKING, List, Union
 
 from coincurve import PublicKeyXOnly
 
 from .crypto.secp import PrivateKey, PublicKey
 from .errors import InvalidProofsError
 from .secret import Secret, SecretKind
 
+if TYPE_CHECKING:
```
a1denvalu3: ?

SatsAndSports: Found a fix, i.e. removing the `TYPE_CHECKING` import and `if` entirely. **But I'm curious what you think before pushing the fix**

_TL:DR it's about an `import` that was only used by `ruff` and not actually used at runtime. Therefore, the idea is to avoid this unnecessary-at-runtime import, perhaps making runtime slightly more efficient and perhaps making the code more maintainable by signalling to maintainers which imports are really needed for which purpose_

This PR introduces a new function `def sig_all_swap_message(proofs: List["Proof"], outputs: List["BlindedMessage"]) -> str:`. I'll double check if this new function is necessary; maybe it's redundant in this draft PR

_(With the fix, we can remove the `"` from that signature and simply use `List[Proof]` and `List[BlindedMessage]`. But that's a slightly different issue to the issue described in this comment)_

`ruff` needs an `import` of `Proof` and `BlindedMessage`, otherwise we get this error : `cashu/core/p2pk.py:76:64: F821 Undefined name BlindedMessage`. So that requires importing those two. However, as this import is needed only by `ruff` (and maybe also by `mypy`?) and is not actually used at runtime, we can use the `if TYPE_CHECKING` guard to skip the runtime import.

_**Question**_. `TYPE_CHECKING` allows us to skip imports that aren't needed at runtime, and which are needed only for linting and type-checking. It might make the code more maintainable, by helping readers understand what code is really needed. Fewer runtime deps means a smaller chance of things going run in complex ways at runtime. But it might make things less maintainable for maintainers who don't know `TYPE_CHECKING` (I discovered it only yesterday!). _**Should we leave this as-is, using TYPE_CHECKING like this, or just go with the import at runtime?**_

SatsAndSports: I've removed that `TYPE_CHECKING` stuff. I find it interesting, and tempted to keep it. But it's a broader issue, unrelated to the bug-fixing scope of this PR, and therefore I've removed it

cashu/mint/auth/server.py
```diff
@@ -222,6 +223,15 @@ async def verify_blind_auth(self, blind_auth_token):
...
         try:
             proof = AuthProof.from_base64(blind_auth_token).to_proof()
+            try:
+                Secret.deserialize(proof.secret)
+                logger.warning(
+                    "Blind-auth token uses a NUT-10-style secret, but blind-auth "
+                    "does not enforce spending conditions in this path. Treating it "
+                    "as a plain secret."
+                )
+            except Exception:
```
SatsAndSports: Are auth tokens supposed to follow NUT-10?

If I remember correctly, the existing data structure for an auth proof doesn't have a witness field and therefore I guess it couldn't have been enforced. Therefore, I added this code to make it more explicit.

I couldn't see an easily way to integrate the NUT-10 logic into auth, in the same way that it's integrated into swap and melt.

Should we add something more explicit in the nuts, perhaps adjust NUT-10 to emphasize that it's for swap and melt and not for auth?

a1denvalu3:
@SatsAndSports needs rebase on the latest main

SatsAndSports:
One comment, partly as a reminder to myself - or to anyone else (@a1denvalu3 ?) - that takes this up:

This PR uses `assert` in production. It's for the 'programmer error' case, not the 'user error' case, and therefore I think it might be appropriate for some projects, but maybe that's not the convention in projects like this (@callebtc mentioned this topic before, that we shouldn't use assert like this). An example is [`cashu/mint/conditions.py:114`](https://github.com/SatsAndSports/nutshell/blob/86630d3d6f2c4c08ffc7f84be51fc533c1b7f4a9/cashu/mint/conditions.py#L114) where there is a private function for handling a SIG_ALL case, and that function should be called only when it has already been established that there is at least one SIG_ALL proof in the set of proofs. If the assertion fails, it does _not_ represent a 'normal' error; it's a fundamental breach of the internal contract of how the nutshell functions call each other. Anyway, I'll stop rambling 😀 : **Q: What should we do in these cases? Reject the mint/swap and log a clear error and ask the mint operator to report this an issue? What error do we return to the wallet?**

PR #1010, opened 2026-05-19
feat(crypto): deterministic DLEQ nonce derivation (NUT-12)

robwoodgate:
Implements the deterministic nonce derivation from cashubtc/nuts#368.

## Summary

Derives the DLEQ nonce `r` deterministically from the mint's private key and the proof context, instead of sourcing it from the RNG:

```
r = HMAC-SHA256(key=a, "Cashu_DLEQ_R_v1" || A || B' || C' || ctr)
```

`a` is the 32-byte secp256k1 private key scalar (big-endian); `A`, `B'`, `C'` are uncompressed SEC1 points (65 bytes each); `ctr` is a single byte starting at `0x00`, incremented if the reduced value is `0` or >= n (max 256 attempts).

This removes the dependency on RNG quality. Reusing `r` across two proofs with different challenges leaks the private key immediately via `a = (s₁ - s₂)·(e₁ - e₂)⁻¹ mod n`.

## Changes

- **`cashu/core/crypto/b_dhke.py`** — new `derive_dleq_nonce(a, A, B_, C_)`. `step2_bob_dleq` computes `C_`/`A` first, then derives `p` deterministically. The `p_bytes` override is retained for existing test vectors.
- **`tests/test_crypto.py`** — adds the NUT-12 deterministic-nonce test vector (asserts every fixed value plus `e == hash(R1, R2, A, C_)` verification), reproducibility, and context-sensitivity tests.

## Notes

- `step2_bob` (used by the mint in `ledger.py`) now emits deterministic nonces for every signature. DLEQ verification is unaffected — `(e, s)` remain valid proofs. No on-disk format or API change.
- Cross-checked against the canonical cashu-ts NUT12 implementation; the Python output reproduces `e`, `s`, `A`, and `C_` from the spec vector exactly.

## Test plan

- `tests/test_crypto.py` — 19 passed, including the new NUT-12 vector and the two new determinism tests.
- `make check` (ruff + mypy) passes.

### DIFF
--- a/cashu/core/crypto/b_dhke.py
+++ b/cashu/core/crypto/b_dhke.py
@@ -160,8 +160,7 @@ def derive_dleq_nonce(
         r = x - SECP256K1_N if x >= SECP256K1_N else x
         if r != 0:
             return PrivateKey(r.to_bytes(32, "big"))
-    # Practically unreachable
-    raise ValueError("DLEQ nonce derivation failed")
+    raise ValueError("DLEQ nonce derivation failed")  # pragma: no cover
 
 
 def step2_bob_dleq(
--- a/cashu/core/crypto/b_dhke.py
+++ b/cashu/core/crypto/b_dhke.py
@@ -154,11 +154,11 @@ def derive_dleq_nonce(
     )
     for ctr in range(256):
         h = hmac.new(a.secret, base + bytes([ctr]), hashlib.sha256).digest()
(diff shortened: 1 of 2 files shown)


robwoodgate:
@a1denvalu3 - have updated this to use rejection sampling as per change in the NUT

prusnak (approved):

callebtc (approved):

PR #1011, opened 2026-05-19
Avoid Postgres test fixture deadlocks

SatsAndSports:
While working on https://github.com/cashubtc/nutshell/pull/1008 , I found that some existing tests would hang locally.

This is a test-only change.

This PR avoids deadlocks in the function-scoped ledger test fixture when running the suite against Postgres. The fixture previously called ledger.startup_ledger(), which starts background tasks such as the regular maintenance loop and invoice listeners. In Postgres-backed test runs, those long-lived tasks could keep transactions open while the next test tried to reset the database with DROP SCHEMA public CASCADE, causing the suite to hang. With this PR, instead only the minimum needed functionality - no background task loops - will be started.

===

With postgres running, I run this `MINT_BACKEND_BOLT11_SAT=FakeWallet WALLET_NAME=test_wallet MINT_HOST=localhost MINT_PORT=3337 MINT_TEST_DATABASE=postgres://cashu:cashu@localhost:5432/cashu DEBUG_MINT_ONLY_DEPRECATED=false TOR=false poetry run pytest tests/mint/test_mint.py -vv -s`

Without this fix, it always hangs - usually on `tests/mint/test_mint.py::test_keysets`. With this fix, it never hangs

### DIFF
--- a/.github/workflows/regtest-mint.yml
+++ b/.github/workflows/regtest-mint.yml
@@ -61,6 +61,7 @@ jobs:
           # LndRestWallet
           MINT_LND_REST_ENDPOINT: https://localhost:8081/
           MINT_LND_REST_CERT: ./regtest/data/lnd-3/tls.cert
+          MINT_LND_REST_CERT_VERIFY: false
           MINT_LND_REST_MACAROON: ./regtest/data/lnd-3/data/chain/bitcoin/regtest/admin.macaroon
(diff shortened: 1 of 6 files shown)


SatsAndSports:
A small fix that helps me running tests on my laptop, might be relevant in CI too

( @a1denvalu3 )

PR #1016, opened 2026-05-25
fix(mint): delete orphan blank outputs on melt change early-return

gorrdy:
## Summary
  
  - `_generate_change_promises` deletes the wallet's blank NUT-08 outputs
  from the `promises` table after signing them, but the early-return
  branch (`overpaid_fee <= 0` or no outputs supplied) skipped that
  cleanup. Those rows were left behind with `c_ IS NULL` and collided with
   later operations that re-derive the same `B_` — most visibly NUT-13
  seed restore, which would surface as `OutputsArePendingError`.
  - Apply the same `delete_blinded_messages_melt_id` cleanup in the
  early-return branch.
  - Soften the `overpaid_fee < 0` log from `error` to `debug` with a
  clearer message. In practice this is reachable whenever the LN backend
  reports a fee greater than the wallet-provided reserve (e.g. an LNbits
  backend skimming a service fee on top of the routing fee), so the
  original "this should not happen" wording is misleading and the
  error-level severity is noisy.
  - Add a parametrized regression test covering both `overpaid_fee == 0`
  and `overpaid_fee < 0` cases.
  
  ## Test plan
  
  - [ ] `poetry run pytest tests/mint/test_mint_melt.py -k early_return
  -v` passes both parametrize cases (`overpaid_fee_zero`,
  `overpaid_fee_negative`).
  - [ ] Same tests fail on `main` (without the ledger fix) — orphan rows
  assertion trips. 
  - [ ] Full melt suite still green: `poetry run pytest
  tests/mint/test_mint_melt.py`.

### DIFF
--- a/cashu/mint/ledger.py
+++ b/cashu/mint/ledger.py
@@ -270,10 +270,6 @@ class Ledger(
                     f"No change to return: backend fee {fee_paid} exceeds wallet's "
                     f"fee reserve {fee_provided} by {-overpaid_fee}."
                 )
-            # Clean up the blank outputs the wallet sent for fee return; otherwise
-            # they remain in `promises` with c_ IS NULL and collide with later
-            # operations that re-derive the same B_ (e.g. NUT-13 seed restore).
-            # Skip when outputs is None — nothing was stored in the first place.
             if melt_id and outputs is not None:
                 await self.crud.delete_blinded_messages_melt_id(
                     melt_id=melt_id, db=self.db
--- a/cashu/mint/ledger.py
+++ b/cashu/mint/ledger.py
@@ -266,8 +266,16 @@ class Ledger(
 
         if overpaid_fee <= 0 or outputs is None:
             if overpaid_fee < 0:
-                logger.error(
-                    f"Overpaid fee is negative ({overpaid_fee}). This should not happen."
+                logger.debug(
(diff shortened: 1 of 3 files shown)


a1denvalu3:
looks like the CI is failing

KvngMikey:
@gorrdy , you've got this test failing:

`FAILED tests/mint/test_mint.py::test_generate_change_promises_zero_fee_deletes_all_blanks - assert 0 == 4
 +  where 0 = len([])`

cashu/mint/ledger.py
```diff
@@ -253,8 +253,16 @@ async def _generate_change_promises(
...
-                logger.error(
-                    f"Overpaid fee is negative ({overpaid_fee}). This should not happen."
+                logger.debug(
+                    f"No change to return: backend fee {fee_paid} exceeds wallet's "
+                    f"fee reserve {fee_provided} by {-overpaid_fee}."
+                )
+            # Clean up the blank outputs the wallet sent for fee return; otherwise
+            # they remain in `promises` with c_ IS NULL and collide with later
+            # operations that re-derive the same B_ (e.g. NUT-13 seed restore).
+            if melt_id:
```
KvngMikey: if melt_id and outputs is not None:  # only clean up if something was stored

gorrdy:
Should be ok now.

KvngMikey (approved):

PR #1022, opened 2026-05-27
fix(wallet): handle unpaid invoices in the CLI invoice command

### LINKED ISSUE #1021: [Wallet] CLI invoice command: misleading "Invoice paid" output + indefinite websocket hang on unpaid invoices
Currently the code doesn't handle the case where an invoice initiated from the CLI isn't paid. The output in the terminal still says "Invoice Paid" for cases where the mint doesn't support websockets and where it does, it hangs indefinitely.

**Reproduction**
Before running the server:
1. To reproduce this while testing in the cli via FakeWallet, you can turn off auto invoice payment by setting  `fakewallet_brr=False` in the settings file. ( cashu/core/settings.py ) under the FakeWalletSettings class.

     This will remove the automatic payment of the invoice generated by the cli
2. Run the server
3. Run the command `cashu invoice 99` or any other amount
    If the mint supports websockets, the code will hang in this code block ( in cashu/wallet/cli/cli.py ) while waiting for the invoice to be paid
    ```python
    if mint_supports_websockets:
        while not paid:
            await asyncio.sleep(0.1)
    ```
    If the mint doesn't support websockets, then the code polls the invoice state for 5 minutes after which this is executed
    ```python
    if not paid:
        print("\n")
        print(
                  "Invoice is not paid yet, stopping check. Use the command above to"
                  " recheck after the invoice has been paid."
                )
    ```
    But after this, it simply falls through to the end of the method and then "Invoice Paid" is printed despite the invoice not being paid
    ```python
    # close open subscriptions so we can exit
        try:
            subscription.close()
        except Exception:
            pass
        print(" Invoice paid.")
    
        print("")
        await print_balance(ctx)
        return
    ```
    It's also worth noting that the `mint_quote` object already has an `expiry` field but neither execution path utilizes it.

**Test Gap**

I noticed that the tests also don't have a pipeline to focusing on if an invoice is not paid. I have not yet thoroughly investigated the testing aspect of it, but the indefinite websocket hang is the type of thing that could cause some tests to hang. Highlighted in #952 

My suspicions are also led by the fact that I experienced a the same while running the tests and the test file that hung was `test_wallet_cli.py` - which tests the particular code that this issue is focused on.

edgarmuyomba:
## Summary
- Added timeout to websocket wait loop using mint quote expiry, preventing indefinite hang when invoice is never paid
- Put "Invoice paid" output behind paid state check so it only prints when minting actually succeeded

## Testing
Tested by setting `fakewallet_brr=False` to disable auto invoice payment, then running `cashu invoice 99` and waiting ~10 minutes.
Console correctly output "Invoice is not paid yet, stopping check..." with no misleading "Invoice paid" message.

### DIFF
--- a/cashu/wallet/cli/cli.py
+++ b/cashu/wallet/cli/cli.py
@@ -600,7 +600,8 @@ async def invoice(
                 flush=True,
             )
         if mint_supports_websockets:
-            while not paid:
+            ws_deadline = mint_quote.expiry or (time.time() + 5 * 60)  # wait for five minutes
+            while not paid and time.time() < ws_deadline:
                 await asyncio.sleep(0.1)
 
         # we still check manually every 10 seconds
@@ -648,10 +649,13 @@ async def invoice(
         subscription.close()
     except Exception:
         pass
-    print(" Invoice paid.")
 
-    print("")
-    await print_balance(ctx)
+    if paid:
+        print(" Invoice paid.")
+
+        print("")
+        await print_balance(ctx)
+
     return
 
 


b-l-u-e:
Faced similar issue 
Nice PR will review it today

b-l-u-e:
Tested locally and confirmed

- `cashu invoice` no longer blocks indefinitely in the websocket wait loop when the quote stays unpaid (bounded by quote `expiry` or the existing 5‑minute fallback).
- It no longer prints misleading `Invoice paid.` when payment never completed.
- The wallet calls `subscription.close()` and exits cleanly after stopping the check.

Test setup: local mint with websocket support, unpaid invoice (e.g. `FAKEWALLET_BRR=false`), `cashu invoice <amount>` with websocket polling enabled.

callebtc (approved):
thank you, awesome PR. sorry for the delay!

PR #1029, opened 2026-05-30
fix(lightning): enforce max fee reserve on blink payment quotes

a1denvalu3:
## Description

This PR fixes a potential vulnerability in the Blink lightning backend where the returned routing fee from the Blink API was blindly trusted without any maximum bounds. An attacker could exploit this by posing as a malicious node in the routing path and returning an extremely high fee during the probe. 

The fix enforces a max fee reserve by capping the backend's API fee probe with the mint's global `fee_reserve`. It ensures the backend never proposes a fee unbounded beyond the system limits or Blink's defined default minimums (`0.5%` or `10 sats`).

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Verified by successfully running `make check` and the Blink-specific test suite `pytest tests/mint/test_mint_lightning_blink.py`.

### DIFF
--- a/cashu/lightning/blink.py
+++ b/cashu/lightning/blink.py
@@ -592,11 +592,24 @@ class BlinkWallet(LightningBackend):
             math.ceil(amount_msat / 100 * BLINK_MAX_FEE_PERCENT / 1000) * 1000
         )
 
+        # The absolute maximum fee we are willing to accept from the probe.
+        # We must respect Blink's hardcoded minimums (10 sats or 0.5%) for small payments,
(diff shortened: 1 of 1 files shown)


a1denvalu3:
I've updated the PR to throw an exception if the probed fee exceeds the `max_allowed_fee_msat` instead of capping it.

As discussed, since Blink's `pay_invoice` doesn't support a fee limit parameter, capping the fee in the quote would cause the mint to effectively subsidize malicious routing fees out of its own pocket. Throwing an exception properly aborts the quote process for malicious invoices.

PR #1031, opened 2026-05-31
fix(mint): enforce auth for NUT-29 batch minting endpoint

a1denvalu3:
Enforce blind authentication for the NUT-29 batch minting endpoint (`/v1/mint/bolt11/batch`) when `settings.mint_require_auth` is enabled.


KvngMikey (approved):

KvngMikey:
ACK

PR #1036, opened 2026-06-03
fix(mint): close websocket when subscribed mint quotes expire unpaid

b-l-u-e:
Inspired by the quote lifecycle work in #1022 (wallet-side handling of unpaid/expired invoices), this PR hardens the **mint** side: websocket subscriptions to `bolt11_mint_quote` are closed when all subscribed quotes are **unpaid** and **past `expiry`**, instead of holding connections open until `mint_websocket_read_timeout` (default 600s).

Also fixes mint quote `expiry` not being persisted or returned on GET: POST returned `expiry`, but GET and the websocket monitor saw `null` because `mint_quotes` had no `expiry` column and `store_mint_quote` never wrote it.

### Problem

- WebSocket clients subscribing to an unpaid mint quote could keep a connection open until the read timeout (600s) after the quote TTL, even though the quote is no longer actionable.
- `expiry` was computed at quote creation but not stored in `mint_quotes` (unlike `melt_quotes`), so REST GET, `MintQuote.from_row`, and the expiry monitor could not see it.

### Changes

- **`MintQuote.from_row`**: load `expiry` from DB rows (SQLite + Postgres).
- **`store_mint_quote`**: persist `expiry` on insert.
- **Migration `m036_add_expiry_to_mint_quotes`**: add `expiry` column to `mint_quotes`.
- **`LedgerEventClientManager`**: background poll (`MINT_WEBSOCKET_QUOTE_EXPIRY_CHECK_INTERVAL`, default 30s) closes WS with code `1000` / reason `mint quote subscription expired` when every subscribed mint quote is unpaid and `expiry <= now`.
- **Tests**: websocket monitor unit tests + assert TTL quote survives round-trip via `get_mint_quote`.

###  Verification

Local mint:

```env
MINT_QUOTE_TTL=60
FAKEWALLET_BRR=false
MINT_WEBSOCKET_QUOTE_EXPIRY_CHECK_INTERVAL=5
```

```
export MINT_URL=http://127.0.0.1:3338
QUOTE_RESP=$(curl -sS -X POST "$MINT_URL/v1/mint/quote/bolt11" \
  -H 'Content-Type: application/json' -d '{"amount": 64, "unit": "sat"}')
QUOTE_ID=$(echo "$QUOTE_RESP" | jq -r .quote)
```

**POST and GET both return matching expiry:**

```json
{ "quote": "k4D70Q30zKmxgnAGmxGcGLYSdRNKL71TXj8ZRttb", "state": "UNPAID", "expiry": 1780484602 }
```

**WebSocket**  (python3 /tmp/ws_quote_hold.py "$QUOTE_ID"), ~60s after quote creation:

```
CLOSED 1000 mint quote subscription expired
```

**Mint log:**
```
Closing websocket: all subscribed mint quotes expired unpaid
```

After expiry, GET still returns state: "UNPAID" with expiry set  expected: state tracks payment/mint progress; expiry is a separate deadline. Clients treat UNPAID + expiry < now as dead; POST /v1/mint/bolt11 rejects with quote expired.

### Test plan

```
 pytest tests/mint/test_mint_websocket_protocol.py

 pytest tests/mint/test_mint.py::test_mint_quote_ttl_setting_overrides_invoice_expiry

 Manual repro above (POST/GET expiry match, WS close at TTL + poll interval)
 ```

### DIFF
--- a/cashu/mint/events/client.py
+++ b/cashu/mint/events/client.py
@@ -1,5 +1,7 @@
 import asyncio
+import contextlib
 import json
+import time
 from typing import List, Union
 
 from fastapi import WebSocket, WebSocketDisconnect
@@ -46,6 +48,19 @@ class LedgerEventClientManager:
     async def start(self):
         await self.websocket.accept()
 
+        # Close connections that only subscribe to mint quotes which expire
+        # unpaid, so a client cannot hold the socket open until the read timeout.
+        expiry_monitor_task = asyncio.create_task(
+            self._monitor_expired_mint_quote_subscriptions()
+        )
+        try:
+            await self._receive_loop()
+        finally:
+            expiry_monitor_task.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await expiry_monitor_task
+
+    async def _receive_loop(self):
         while True:
             message = await asyncio.wait_for(
                 self.websocket.receive(),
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: cashu/mint/events/client.py [hunk 1]
```diff
@@ -174,6 +189,55 @@ async def _send_msg(
...
+                    # Keep the connection alive if any subscribed quote is
+                    # missing expiry data, is not unpaid, or has not expired yet.
+                    if (
+                        not mint_quote
+                        or not mint_quote.unpaid
+                        or not mint_quote.expiry
+                        or mint_quote.expiry > now
+                    ):
+                        all_expired = False
+                        break
```
edgarmuyomba: Correct me if I'm wrong, looking at the condition — a paid quote causes `not mint_quote.unpaid` to be true, which sets `all_expired = False` and breaks. So a paid quote actually keeps the connection alive. If a wallet doesn't close the connection after payment (even though nutshell's wallet does), this task won't clean it up either. 

Would it make sense to treat paid quotes as terminal too, so the check becomes "close if every subscribed quote is either paid or expired+unpaid"?

tests/mint/test_mint_websocket_protocol.py
```diff
@@ -282,3 +292,104 @@ async def test_event_manager_rejects_unsupported_events():
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "state, expiry",
+    [
+        # unpaid but not yet expired -> keep open
+        (MintQuoteState.unpaid, lambda: int(time.time()) + 3600),
+        # already paid (even if expiry is in the past) -> keep open
+        (MintQuoteState.paid, lambda: int(time.time()) - 10),
```
edgarmuyomba: Referring to earlier suggestion about Paid being a terminal state so perhaps it should also be closed.

cashu/mint/events/client.py
callebtc: I think that makes sense. 

afaik we don't respect the expired field for mint quotes in the mint, so I'm not sure if we should drop the connection

b-l-u-e:
Thank you all for the feedback

made changes now we close WS once all subscribed mint quotes are terminal (`PAID` or `UNPAID && expiry <= now`).  
Updated monitor logic + tests (including replacing the old “paid keeps open” case).

callebtc:
I may be not the best idea to enforce restrictions on the websocket as long as we don't restrict anything in the business logic (of minting itself) that depends on the expiry. We don't seem to have consensus on what to do with the expiry yet.

edgarmuyomba:
> I may be not the best idea to enforce restrictions on the websocket as long as we don't restrict anything in the business logic (of minting itself) that depends on the expiry. We don't seem to have consensus on what to do with the expiry yet.

Valid concern — makes sense to wait for consensus on expiry semantics before enforcing them here. Though I think closing connections once all subscribed quotes are paid is worth doing on its own, because a subscribed connection to an already-minted quote has no server-side cleanup mechanism. Also, scoping to just the paid case (for now) would leave the monitor structure in place, making it straightforward to add expiry-based handling once that's settled.

PR #1050, opened 2026-06-12
fix(lightning): migrate LND backends to SendPaymentV2 (SendPaymentSync removed in lnd 0.21)

### LINKED ISSUE #1049: [BUG] Melt fails on LND v0.21: lndrest and lnd_grpc backends use SendPaymentSync, removed in lnd v0.21.0-beta
# PR https://github.com/cashubtc/nutshell/pull/1050

## Summary

  LND **v0.21.0-beta** (released 2026-06-05) removed the long-deprecated payment RPCs, including `lnrpc.SendPaymentSync`  and its REST route `POST /v1/channels/transactions`. From the [v0.21.0 release
  notes](https://github.com/lightningnetwork/lnd/blob/master/docs/release-notes/release-notes-0.21.0.md):

  > Removed the deprecated payment RPCs [...] This also removes the corresponding REST routes `POST
  /v1/channels/transaction-stream`, `POST /v1/channels/transactions`, and `POST /v1/channels/transactions/route`.

  Both Nutshell LND backends still pay invoices via this removed RPC, so **every melt fails immediately** against an LND
  0.21 node:

  - `lndrest`: [`cashu/lightning/lndrest.py#L207-L211`](https://github.com/cashubtc/nutshell/blob/0.20.1/cashu/lightning/ln
  drest.py#L207-L211) — `POST /v1/channels/transactions`
  - `lnd_grpc`: [`cashu/lightning/lnd_grpc/lnd_grpc.py#L179`](https://github.com/cashubtc/nutshell/blob/0.20.1/cashu/lightn
  ing/lnd_grpc/lnd_grpc.py#L179) — `lnstub.SendPaymentSync(...)`

  Minting/receiving is unaffected (`POST /v1/invoices` still exists), which makes the failure look like a wallet-side or
  routing problem at first.

  ## Affected versions

  - Nutshell `0.20.1` and all earlier releases (the payment path is unchanged for years)
  - LND `v0.21.0-beta` and later

  ## Symptoms / logs

  Reported by a mint operator (LND REST backend) after their node updated to lnd 0.21 — all melts fail in ~60 ms:

  ```
  DEBUG | cashu.mint.ledger:melt:1106 | Lightning: pay invoice lnbc500n1...
  DEBUG | cashu.mint.ledger:melt:1111 | Melt – Result: FAILED: preimage: None, fee: None
  DEBUG | cashu.mint.ledger:melt:1135 | Payment state is FAILED. Error: {"code":5,"message":"Not Found","details":[]}.
  Checking status for 3774be4f...
  ERROR | cashu.lightning.lndrest:get_payment_status:384 | LND get_payment_status error: payment isn't initiated
  ERROR | cashu.mint.app:catch_exceptions:99 | CashuError: Lightning payment failed: {"code":5,"message":"Not
  Found","details":[]}.
  INFO  | "POST /v1/melt/bolt11 HTTP/1.1" 400
  ```

  Every line is consistent with the removed endpoint:

  - `{"code":5,"message":"Not Found","details":[]}` is the grpc-gateway JSON body for an HTTP 404 on an unregistered REST
  route; `pay_invoice` surfaces the raw body as the payment error.
  - The failure comes back in ~60 ms — no payment attempt ever reaches the LND payment engine.
  - The follow-up status check (`TrackPaymentV2`, which *is* a V2 call and still works) then correctly reports `payment
  isn't initiated`, because LND has no record of the payment.
  - Direct node-to-node test payments work fine, since `lncli` uses `SendPaymentV2`.

  ## What is NOT affected

  - Invoice creation (`POST /v1/invoices` / `AddInvoice`)
  - `get_invoice_status` and `get_payment_status` (already on `TrackPaymentV2`)
  - The MPP path `pay_partial_invoice` (already on `SendToRouteV2` / `POST /v2/router/route/send`)

  So the migration surface is just the two `pay_invoice` implementations.

  ## Suggested fix

  Migrate `pay_invoice` to `routerrpc.SendPaymentV2`:

  - REST: `POST /v2/router/send` — note this is a **streaming** endpoint, so the response needs to be consumed line-by-line
  until a terminal payment status (`SUCCEEDED` / `FAILED`), same pattern as the existing `get_payment_status` stream
  handling.
  - gRPC: `router_stub.SendPaymentV2(routerrpc.SendPaymentRequest(...))` — also a stream of `lnrpc.Payment` updates; the
  router stub is already imported and used in `lnd_grpc.py`.
  - `SendPaymentV2` requires `timeout_seconds` (no default); `fee_limit_msat` can be passed directly instead of the
  `FeeLimit` message.
  - The terminal `Payment` statuses map onto the existing `PAYMENT_RESULT_MAP` already used by `get_payment_status`.

  ## Workaround for operators

  Stay on LND `v0.20.x` until a fixed Nutshell release is out.

robwoodgate:
# Fixes #1049

## Problem

LND **v0.21.0-beta** (released 2026-06-05) removed the long-deprecated payment RPCs, including `lnrpc.SendPaymentSync` and its REST route `POST /v1/channels/transactions` ([v0.21.0 release notes](https://github.com/lightningnetwork/lnd/blob/master/docs/release-notes/release-notes-0.21.0.md)):

> Removed the deprecated payment RPCs [...] This also removes the corresponding REST routes `POST /v1/channels/transaction-stream`, `POST /v1/channels/transactions`, and `POST /v1/channels/transactions/route`.

Both Nutshell LND backends paid invoices via this removed RPC, so **every melt fails immediately** against an lnd 0.21 node. Reported by a mint operator (LND REST backend) whose node updated to lnd 0.21 — all melts fail in ~60 ms:

```
DEBUG | cashu.mint.ledger:melt:1111 | Melt – Result: FAILED: preimage: None, fee: None
DEBUG | cashu.mint.ledger:melt:1135 | Payment state is FAILED. Error: {"code":5,"message":"Not Found","details":[]}. Checking status for 3774be4f...
ERROR | cashu.lightning.lndrest:get_payment_status:384 | LND get_payment_status error: payment isn't initiated
ERROR | cashu.mint.app:catch_exceptions:99 | CashuError: Lightning payment failed: {"code":5,"message":"Not Found","details":[]}.
```

`{"code":5,"message":"Not Found"}` is the grpc-gateway body for a 404 on the now-unregistered route; the payment never reaches the payment engine, and the follow-up `TrackPaymentV2` status check correctly reports `payment isn't initiated`. Minting/receiving is unaffected (`POST /v1/invoices` still exists), which makes this look like a wallet-side or routing problem at first.

## Fix

Migrate `pay_invoice` in both backends to the streaming `routerrpc.SendPaymentV2` (available since lnd 0.10, so no minimum-version bump in practice):

- **lndrest**: `POST /v2/router/send` with `no_inflight_updates: true`, consuming the JSON-line stream until a terminal payment status, mirroring the existing `get_payment_status` (`TrackPaymentV2`) handling. Streaming `{"error": ...}` lines and immediate REST-proxy error bodies (`{"code": ..., "message": ...}`) map to `FAILED`; on `FAILED` the `failure_reason` (e.g. `FAILURE_REASON_NO_ROUTE`) is surfaced as the error message.
- **lnd_grpc**: `router_stub.SendPaymentV2(routerrpc.SendPaymentRequest(...))`, same terminal-status handling; `AioRpcError` maps to `FAILED` (matching the previous `SendPaymentSync` behavior).
- If the stream ends without a terminal status, `UNKNOWN` is returned so the ledger's existing status double-check (`get_payment_status` → `TrackPaymentV2`) resolves the payment state — the ledger already re-checks `FAILED`/`UNKNOWN` results before rolling back, so fund-safety semantics are unchanged.
- `timeout_seconds` is required by `SendPaymentV2`; set to 60s (same as the `lncli` default).

Not affected (already on non-removed endpoints): invoice creation (`AddInvoice`), `get_invoice_status`, `get_payment_status` (`TrackPaymentV2`), and the MPP path `pay_partial_invoice` (`SendToRouteV2`).

## Tests

- Updated the mocked lndrest `pay_invoice` test (previously mocked the removed `SendPaymentSync` response shape) and added cases for: settled stream result (fee/preimage/checking_id parsing), terminal `FAILED` with `failure_reason`, immediate REST-proxy error body (the exact lnd 0.21 failure mode), and an empty stream returning `UNKNOWN`.
- `pytest tests/lightning/` passes (31 tests), `ruff check` and `mypy --check-untyped-defs` clean on the changed files.

## Workaround for operators until release

Stay on LND v0.20.x. for now

### DIFF
--- a/cashu/lightning/lnd_grpc/lnd_grpc.py
+++ b/cashu/lightning/lnd_grpc/lnd_grpc.py
@@ -211,9 +211,12 @@ class LndRPCWallet(LightningBackend):
                         ),
                     )
         except AioRpcError as e:
+            # transport/RPC error: we don't know whether the payment was
+            # attempted, so report UNKNOWN (never FAILED) and let the ledger
+            # re-check the real state with TrackPaymentV2.
             error_message = f"SendPaymentV2 failed: {e}"
             return PaymentResponse(
-                result=PaymentResult.FAILED,
+                result=PaymentResult.UNKNOWN,
                 error_message=error_message,
             )
 
--- a/cashu/lightning/lnd_grpc/lnd_grpc.py
+++ b/cashu/lightning/lnd_grpc/lnd_grpc.py
@@ -47,6 +47,7 @@ INVOICE_RESULT_MAP = {
 }
 
 MAX_ROUTE_RETRIES = 50
+PAYMENT_TIMEOUT_SECONDS = 60
 
 
 class LndRPCWallet(LightningBackend):
@@ -168,42 +169,59 @@ class LndRPCWallet(LightningBackend):
                     quote, quote_amount, fee_limit_msat
                 )
 
-        # set the fee limit for the payment
-        feelimit = lnrpc.FeeLimit(fixed_msat=fee_limit_msat)
-        r = None
+        # pay the invoice with routerrpc.SendPaymentV2, a streaming RPC that
+        # sends payment updates until a terminal status is reached. The
+        # previously used lnrpc.SendPaymentSync was removed in lnd v0.21.0.
+        request = routerrpc.SendPaymentRequest(
+            payment_request=quote.request,
+            fee_limit_msat=fee_limit_msat,
+            timeout_seconds=PAYMENT_TIMEOUT_SECONDS,
+            no_inflight_updates=True,
+        )
         try:
             async with grpc.aio.secure_channel(
                 self.endpoint, self.combined_creds
             ) as channel:
-                lnstub = lightningstub.LightningStub(channel)
-                r = await lnstub.SendPaymentSync(
-                    lnrpc.SendRequest(
-                        payment_request=quote.request,
-                        fee_limit=feelimit,
+                router_stub = routerstub.RouterStub(channel)
+                async for payment in router_stub.SendPaymentV2(request):
+                    result = PAYMENT_RESULT_MAP.get(
+                        payment.status, PaymentResult.UNKNOWN
                     )
-                )
+                    if result == PaymentResult.PENDING:
+                        # non-terminal in-flight update, wait for the next one
+                        continue
+
+                    if result == PaymentResult.SETTLED:
+                        return PaymentResponse(
+                            result=PaymentResult.SETTLED,
+                            checking_id=payment.payment_hash,
+                            fee=(
+                                Amount(unit=Unit.msat, amount=payment.fee_msat)
+                                if payment.fee_msat
+                                else None
+                            ),
+                            preimage=payment.payment_preimage,
+                            error_message=None,
+                        )
+
+                    return PaymentResponse(
+                        result=result,
+                        error_message=lnrpc.PaymentFailureReason.Name(
+                            payment.failure_reason
+                        ),
+                    )
         except AioRpcError as e:
-            error_message = f"SendPaymentSync failed: {e}"
+            error_message = f"SendPaymentV2 failed: {e}"
             return PaymentResponse(
                 result=PaymentResult.FAILED,
                 error_message=error_message,
             )
 
-        if r.payment_error:
-            return PaymentResponse(
-                result=PaymentResult.FAILED,
-                error_message=r.payment_error,
-            )
-
-        checking_id = r.payment_hash.hex()
-        fee_msat = r.payment_route.total_fees_msat
-        preimage = r.payment_preimage.hex()
+        # stream ended without a terminal status, get_payment_status will
+        # check the payment state with TrackPaymentV2
         return PaymentResponse(
-            result=PaymentResult.SETTLED,
-            checking_id=checking_id,
-            fee=Amount(unit=Unit.msat, amount=fee_msat) if fee_msat else None,
-            preimage=preimage,
-            error_message=None,
+            result=PaymentResult.UNKNOWN,
+            error_message="SendPaymentV2 stream ended without a terminal payment status",
         )
 
     async def pay_partial_invoice(
--- a/cashu/lightning/lndrest.py
+++ b/cashu/lightning/lndrest.py
@@ -221,18 +221,25 @@ class LndRestWallet(LightningBackend):
                 except json.JSONDecodeError:
                     continue
 
-                # streaming errors from the REST proxy
+                # streaming errors from the REST proxy: the stream errored
+                # after the request was accepted (e.g. gRPC AlreadyExists when
+                # the payment is already in flight). We don't know whether the
+                # payment is live, so report UNKNOWN (never FAILED) and let the
+                # ledger re-check the real state with TrackPaymentV2.
                 if line.get("error"):
                     error = line["error"]
                     message = (
                         error["message"] if "message" in error else str(error)
                     )
                     return PaymentResponse(
-                        result=PaymentResult.FAILED, error_message=message
+                        result=PaymentResult.UNKNOWN, error_message=message
                     )
 
-                # immediate (non-streaming) errors from the REST proxy,
-                # e.g. {"code": 5, "message": "Not Found", "details": []}
+                # immediate (non-streaming) errors from the REST proxy reject
+                # the request before any payment is initiated, e.g.
+                # {"code": 5, "message": "Not Found", "details": []} when the
+                # route is unavailable. No payment was created, so this is an
+                # explicit FAILED.
                 if line.get("message") and "result" not in line:
                     return PaymentResponse(
                         result=PaymentResult.FAILED, error_message=line["message"]
--- a/cashu/lightning/lndrest.py
+++ b/cashu/lightning/lndrest.py
@@ -43,6 +43,7 @@ INVOICE_RESULT_MAP = {
 
 MAX_ROUTE_RETRIES = 50
 TEMPORARY_CHANNEL_FAILURE_ERROR = "TEMPORARY_CHANNEL_FAILURE"
+PAYMENT_TIMEOUT_SECONDS = 60
 
 
 class LndRestWallet(LightningBackend):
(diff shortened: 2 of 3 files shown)


cashu/lightning/lnd_grpc/lnd_grpc.py
```diff
@@ -47,6 +47,7 @@
 }
 
 MAX_ROUTE_RETRIES = 50
+PAYMENT_TIMEOUT_SECONDS = 60
```
a1denvalu3, line 50: IIRC we specifically set the timeout to None.

cashu/lightning/lnd_grpc/lnd_grpc.py
```diff
@@ -168,42 +169,59 @@ async def pay_invoice(
...
-        feelimit = lnrpc.FeeLimit(fixed_msat=fee_limit_msat)
-        r = None
+        # pay the invoice with routerrpc.SendPaymentV2, a streaming RPC that
+        # sends payment updates until a terminal status is reached. The
+        # previously used lnrpc.SendPaymentSync was removed in lnd v0.21.0.
+        request = routerrpc.SendPaymentRequest(
+            payment_request=quote.request,
+            fee_limit_msat=fee_limit_msat,
+            timeout_seconds=PAYMENT_TIMEOUT_SECONDS,
+            no_inflight_updates=True,
```
a1denvalu3, line 179: What does `no_inflight_updates` do?

cashu/lightning/lnd_grpc/lnd_grpc.py
```diff
@@ -168,42 +169,59 @@ async def pay_invoice(
...
-                lnstub = lightningstub.LightningStub(channel)
-                r = await lnstub.SendPaymentSync(
-                    lnrpc.SendRequest(
-                        payment_request=quote.request,
-                        fee_limit=feelimit,
+                router_stub = routerstub.RouterStub(channel)
+                async for payment in router_stub.SendPaymentV2(request):
+                    result = PAYMENT_RESULT_MAP.get(
+                        payment.status, PaymentResult.UNKNOWN
+                    )
```
a1denvalu3, line 189: As a rule of thumb, once we send the payment we always assign PENDING unless the backend explicitly tells us it failed. Even on network errors or weird things.

cashu/lightning/lnd_grpc/lnd_grpc.py
```diff
@@ -168,42 +169,59 @@ async def pay_invoice(
...
-        preimage = r.payment_preimage.hex()
+        # stream ended without a terminal status, get_payment_status will
+        # check the payment state with TrackPaymentV2
         return PaymentResponse(
-            result=PaymentResult.SETTLED,
-            checking_id=checking_id,
-            fee=Amount(unit=Unit.msat, amount=fee_msat) if fee_msat else None,
-            preimage=preimage,
-            error_message=None,
+            result=PaymentResult.UNKNOWN,
```
a1denvalu3, line 226: As a rule of thumb, once we send the payment we always assign PENDING unless the backend explicitly tells us it failed. Even on network errors or weird things.

a1denvalu3 (dismissed):
@callebtc We should also probably update the cashu_regtest with the latest LND.

cashu/lightning/lnd_grpc/lnd_grpc.py
robwoodgate: I think that's the transport timeout (httpx/gRPC client deadline) you set to None. That is preserved: lndrest.py keeps timeout=None on the stream call, and the gRPC path sets no client deadline.

This constant sets `timeout_seconds`, which is a field inside `SendPaymentRequest`. If I understand correctly, it's how long LND itself searches for a route before giving up. This is mandatory in SendPaymentV2; LND rejects 0. So it genuinely cannot be None.

The bigger decision is "how long should it be". Is 60 seconds the right value? Happy to set whatever you think it should be.

cashu/lightning/lnd_grpc/lnd_grpc.py
robwoodgate: The setting tells LND not to stream intermediate states, only the terminal payment result. 

The loop  acts on terminal states, so this setting cuts stream noise. As a side effect: the PENDING → continue branch becomes essentially defensive/dead code (harmless).

cashu/lightning/lnd_grpc/lnd_grpc.py
robwoodgate: The ledger treats FAILED and UNKNOWN identically, both trigger a second, independent status check via TrackPaymentV2 `(ledger.py:1150,1157)`, and it only rolls the quote back to unpaid if that re-check also says failed. SETTLED is the only state trusted without a recheck.

So these UNKNOWN returns are already safe, because they get reverified. UNKNOWN is arguably better than PENDING here, because UNKNOWN triggers that automatic re-check (which can auto-resolve to settled/failed), whereas PENDING leaves the quote stuck pending, needing manual operator action.

This also follows the conventions in the Blink backend.

cashu/lightning/lnd_grpc/lnd_grpc.py
robwoodgate: Same reasoning as the line-189 thread. This is "stream ended without a terminal status," i.e. we don't know the outcome, not an explicit failure, so it returns UNKNOWN.

Your "even on network errors" point was spot on, and highlighted a different bug, I think: the AioRpcError handler returned FAILED on a network error (no payment proof either way).

I've changed that to UNKNOWN in the latest push, so FAILED is now only returned when LND explicitly fails or rejects the request.

I went with UNKNOWN rather than PENDING following the Blink backend's convention, and because in the ledger both UNKNOWN and FAILED trigger a TrackPaymentV2 re-check before anything is rolled back, whereas PENDING short-circuits and leaves the quote stuck pending until it's resolved manually.

cashu/lightning/lnd_grpc/lnd_grpc.py
robwoodgate: I wonder if 300 is actually more appropriate because of multi-path routes, probing etc?

cashu/lightning/lndrest.py
```diff
@@ -221,18 +221,25 @@ async def pay_invoice(
                 except json.JSONDecodeError:
                     continue
 
-                # streaming errors from the REST proxy
+                # streaming errors from the REST proxy: the stream errored
+                # after the request was accepted (e.g. gRPC AlreadyExists when
+                # the payment is already in flight). We don't know whether the
+                # payment is live, so report UNKNOWN (never FAILED) and let the
+                # ledger re-check the real state with TrackPaymentV2.
```
a1denvalu3, line 228: Thinking out loud

a1denvalu3:
Ran the CI with the updated cashu-regtest, which proves the fixes work. Failures involve LNBits, which on the regtest isn't yet updated to the latest image.

PR #1054, opened 2026-06-17
feat(crypto): align NUT-20/29 signatures with spec change

a1denvalu3:
Aligns NUT-20/29 batch minting signatures to the spec changes in cashubtc/nuts pull request 375, while retaining fallback legacy signature verification for older wallets.


a1denvalu3:
This aligns nutshell with the NUT-20/29 spec changes defined in: https://github.com/cashubtc/nuts/pull/375

PR #1058, opened 2026-06-20
feat(mint): introduce automatic keyset rotations

a1denvalu3:
This PR introduces automatic keyset rotations for Nutshell to encourage/force operators to rotate keysets regularly, as operators rarely perform them manually.

### What's Changed
- **Automatic Rotations Configuration**: Added `MINT_KEYSET_ROTATION_ENABLED` (default: `True`) and `MINT_KEYSET_ROTATION_INTERVAL_SECONDS` (default: 30 days) to `cashu/core/settings.py` and documented them in `.env.example`.
- **Automatic Rotation Detection & Logic**:
  - Implemented `should_rotate_keyset` to safely calculate the age of active keysets (handling both sqlite and postgres timestamp differences).
  - Implemented `rotate_keysets_if_needed` to automatically rotate active keysets for all configured units when they exceed the interval, while preserving customized fees, orders, and final expiry.
  - Enhanced `store_keyset` in DB CRUD to update the `valid_from`, `valid_to`, and `first_seen` timestamps on the in-memory keyset object on creation, ensuring memory and DB stay in perfect synchronization.
- **Lifecycle Integration**:
  - Run automatic rotation checks on ledger startup (`_startup_keysets`).
  - Run automatic rotation checks periodically in the regular tasks background loop (`_run_regular_tasks`).
- **Documentation**:
  - Documented automatic keyset rotations behavior and configuration in `README.md`.
  - Added comments explaining that derivation paths are automatically updated in the DB, meaning operators do NOT need to manually change `MINT_DERIVATION_PATH` in `.env` after a rotation.
- **Testing**:
  - Added full integration test suite `tests/mint/test_mint_automatic_rotations.py` covering standard rotation, disabled mode, and parameter preservation.

Passed `ruff check` and `mypy` without any errors.

### DIFF
--- a/cashu/mint/keysets.py
+++ b/cashu/mint/keysets.py
@@ -45,6 +45,7 @@ class LedgerKeysets(SupportsKeysets, SupportsSeed, SupportsDb):
         max_order: Optional[int] = None,
         input_fee_ppk: Optional[int] = None,
         final_expiry: Optional[int] = None,
+        active_keyset_id: Optional[str] = None,
     ) -> MintKeyset:
         """
         This function:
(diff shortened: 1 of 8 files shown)


cashu/mint/keysets.py
```diff
@@ -245,6 +247,48 @@ async def inactivate_base64_keysets(self) -> None:
...
+                    f"Active keyset {keyset.id} for unit {keyset.unit.name} is older than "
+                    f"the configured rotation interval ({settings.mint_keyset_rotation_interval_seconds}s). "
+                    f"Rotating now."
+                )
+                try:
+                    new_keyset = await self.rotate_next_keyset(
+                        unit=keyset.unit,
+                        max_order=len(keyset.amounts),
+                        input_fee_ppk=keyset.input_fee_ppk,
+                        final_expiry=keyset.final_expiry,
```
KvngMikey: nit: do we intend to pass the old keyset final_expiry straight through or we do we want to recompute the final_expiry after rotation ?

KvngMikey (approved):
just the nit, otherwise lgtm!

cashu/mint/ledger.py
```diff
@@ -166,6 +167,7 @@ async def _run_regular_tasks(self) -> None:
         while True:
             try:
                 await self._check_pending_proofs_and_melt_quotes()
+                await self.rotate_keysets_if_needed()
```
KvngMikey, line 171: The effective rotation time is `MINT_KEYSET_ROTATION_INTERVAL_SECONDS + up to MINT_REGULAR_TASKS_INTERVAL_SECONDS`, but the latter (a) isn't in .env.example, (b) defaults to 3600s, and (c) is described only as the invoice-checker interval, with no mention of rotation. 

It's also overloaded, the same tick drives pending_proof checks. I suggest we either document this coupling in the rotation section of .env.example, or decoupling rotation onto its own timer so `ROTATION_INTERVAL_SECONDS` for example means what it says.

callebtc:
Review findings:

1. **High — rotation is not atomic.** [`rotate_next_keyset`](https://github.com/cashubtc/nutshell/blob/ead9cab552306b072f546100a0b48b369215fe06/cashu/mint/keysets.py#L110-L124) stores the new keyset and deactivates the old keyset in separate database transactions. If the second write fails or the process crashes between them, the new keyset is committed while the old one remains active in the database. The automatic wrapper then catches and suppresses the exception ([lines 308–316](https://github.com/cashubtc/nutshell/blob/ead9cab552306b072f546100a0b48b369215fe06/cashu/mint/keysets.py#L308-L316)), so this can surface only later as multiple active/inconsistent keysets. Please make the insert/deactivation one transaction, with an appropriate per-unit lock.

2. **High — concurrent instances/manual rotation can race.** The automatic check has no locking or idempotency. Two mint processes, or the regular task and management RPC, can observe the same stale keyset and rotate concurrently. This can result in duplicate-key failures, divergent `final_expiry`/keyset IDs, or one process continuing to serve a keyset another process has deactivated.

3. **Medium — highest-counter selection is incomplete.** [`rotate_next_keyset`](https://github.com/cashubtc/nutshell/blob/ead9cab552306b072f546100a0b48b369215fe06/cashu/mint/keysets.py#L69-L78) never updates `selected_keyset_counter`. If multiple keysets for one unit are active (possible with multiple derivation paths), automatic rotation may rotate a different keyset than the stale one and leave the stale keyset active. The new loop makes this latent issue relevant because it processes every active keyset ([lines 290–313](https://github.com/cashubtc/nutshell/blob/ead9cab552306b072f546100a0b48b369215fe06/cashu/mint/keysets.py#L290-L313).

Also, rotation is coupled to the shared regular-task loop ([ledger lines 161–174](https://github.com/cashubtc/nutshell/blob/ead9cab552306b072f546100a0b48b369215fe06/cashu/mint/ledger.py#L161-L174)), so it can occur up to the regular-task interval late (default 3600 seconds), and can be delayed by pending-proof/melt processing.

The basic single-process path and the five new rotation tests pass, but I would request changes for atomicity and concurrency safety.

callebtc (changes requested):
lack of atomicity and multiple mint race condition found

callebtc (changes requested):
Review update after the latest fixes:

1. **High — failed rotation still corrupts in-memory state.** `selected_keyset.active` is changed before the transaction commits. If `update_keyset()` or commit fails, the database rolls back but the live object remains inactive. Because automatic rotation suppresses the exception, the mint can be left with zero active in-memory keysets and will not retry. Reproduced with DB active = true, memory active = false, and `should_rotate_keyset()` = false after the simulated failure. The regression test currently checks only the database.

2. **High — other mint instances do not adopt manual rotations.** `rotate_keysets_if_needed()` decides from in-memory state, while database synchronization happens only after a worker has already decided to rotate. A second worker therefore continues advertising and accepting the database-inactive keyset after another worker performs an early manual rotation.

3. **High — automatic rotation still ignores the stale keyset that triggered it.** `active_keyset_id` is used only for the already-inactive check; selection then chooses the greatest counter across every active keyset for the unit. With stale `m/0'/0'/0'` and fresh `m/1'/0'/10'`, repeated checks left the stale branch active while rotating the fresh branch to counters 11 and 12.

4. **Medium — the regular-task scheduling concern remains unresolved.** Rotation is still coupled to pending-proof processing and `MINT_REGULAR_TASKS_INTERVAL_SECONDS`, but the rotation documentation does not disclose the possible delay.

The final-expiry concern is addressed, and the database writes are now transactionally atomic, but the memory rollback and remaining multi-instance issues still require changes.

a1denvalu3:
## Code review: automatic keyset rotations

Reviewed the full diff (\`a935eb9c..987b937d\`). Overall solid work — the atomic rotation under \`db.connect(lock_table="keysets")\`, the concurrent-rotation adoption path, and the regression tests target exactly the right failure modes. All 10 new tests pass, and the existing \`test_mint_keysets.py\` (16 tests) still passes.

### What's good
- **Atomic rotation**: store-new + deactivate-old now commit or roll back together; \`test_regression_non_atomic_rotation\` proves the rollback.
- **Race adoption**: the loser of a rotation race re-syncs from the DB under the lock and adopts the winner's highest-counter keyset, including updating \`self.keyset\`/\`self.derivation_path\`.
- **\`balance\` removal from \`update_keyset\`** is a genuine bug fix — balances are maintained incrementally via \`bump_keyset_balance\` in \`cashu/mint/db/write.py\`, so the old code would clobber the DB-maintained balance with a stale in-memory value on every keyset update.

### Findings

1. **Fail-open on unparseable \`valid_from\` can cause unbounded keyset growth** — \`should_rotate_keyset\` returns \`True\` when parsing fails ("Forcing rotation", \`cashu/mint/keysets.py\`). This is bounded only if newly written timestamps are parseable. If a systemic change (driver upgrade, DB type change) makes *all* \`valid_from\` values unparseable, every regular-tasks tick rotates every active keyset — a new keyset per unit per \`mint_regular_tasks_interval_seconds\`, forever. Recommend failing closed: log an error and skip rotation instead.

2. **\`_parse_valid_from\` relies on naive local-time consistency** — the \`strptime\` fallback and \`db.timestamp_from_seconds\` for Postgres both use server-local time. It round-trips correctly only as long as writer and reader share a timezone. A mint host TZ change shifts every keyset's apparent age, potentially triggering spurious rotations of all keysets at once. Storing/comparing UTC epoch internally would remove this fragility; at minimum it's worth a comment.

3. **\`max_order=len(keyset.amounts)\` silently changes non-standard amount sets** — rotation regenerates amounts as \`[2**i for i in range(max_order)]\`. Preserved for standard power-of-two keysets, but a keyset with custom/non-contiguous amounts would be rotated into a *different* amount set. Probably theoretical today; a guard or comment would document the assumption.

4. **\`store_keyset\` now mutates its argument** — it back-fills \`valid_from\`/\`valid_to\`/\`first_seen\` on the caller's \`MintKeyset\` before insert. Presumably deliberate (keeps the in-memory copy consistent with the DB row, which the in-lock sync relies on), but it's a side effect worth a code comment since the rest of the crud layer treats models as read-only inputs.

5. **Default-on is an operational behavior change** — \`mint_keyset_rotation_enabled=True\` by default means every existing mint silently starts rotating on a 90-day cycle after upgrade, and \`MINT_DERIVATION_PATH\` in \`.env\` becomes advisory. Docs cover it, but given keyset management is security-sensitive, this default deserves explicit maintainer sign-off and a changelog entry.

6. **Minor: tests mutate the shared ledger permanently** — e.g. \`test_automatic_keyset_rotation_flow\` performs a real, DB-persisted rotation of the session-scoped ledger and never restores it. Tests currently pass, but this creates ordering coupling with any future test that hardcodes keyset IDs or derivation counters.

7. **Nit:** the \`return active_keyset\` early-exit doesn't re-check whether the adopted keyset itself is already past the interval — harmless (the next tick catches it), just noting the one-tick delay.

Nothing is a hard blocker except I'd push on finding 1 (fail-open rotation) before merge; the rest are hardening/documentation suggestions.

PR #1059, opened 2026-06-21
feat(core): Updates to nut18 and nut26 payment request

KvngMikey:
implements https://github.com/cashubtc/nuts/pull/381 to keep Nutshell up to date !

### DIFF
--- a/cashu/core/mint_info.py
+++ b/cashu/core/mint_info.py
@@ -3,9 +3,14 @@ from typing import Any, Dict, List, Optional
 
 from pydantic import BaseModel
 
-from .base import Method, Unit
-from .models import MintInfoContact, MintInfoProtectedEndpoint, Nut15MppSupport
-from .nuts.nuts import BLIND_AUTH_NUT, CLEAR_AUTH_NUT, MPP_NUT, WEBSOCKETS_NUT
+from .base import Method, SupportedMethod, Unit
+from .models import (
+    MeltMethodSetting,
+    MintInfoContact,
+    MintInfoProtectedEndpoint,
+    Nut15MppSupport,
+)
+from .nuts.nuts import BLIND_AUTH_NUT, CLEAR_AUTH_NUT, MELT_NUT, MPP_NUT, WEBSOCKETS_NUT
 
 
 def _match_protected_endpoint(endpoint_path: str, request_path: str) -> bool:
@@ -66,6 +71,36 @@ class MintInfo(BaseModel):
                 return True
 
         return False
+
+    def payment_request_method_fee(
+        self, sm: Optional[List[SupportedMethod]], unit: Unit
+    ) -> Optional[int]:
+        """
+        Given the `sm` (supported methods) list from a NUT-18/26 payment request,
+        return the lowest per-method fee this mint owes for melting via one of the
+        listed methods in `unit`, or `None` if it melts none of them.
+        """
+        if not sm:
+            return 0
+        if not self.nuts:
+            return None
+        nut_05 = self.nuts.get(MELT_NUT)
+        if not nut_05 or not nut_05.get("methods"):
+            return None
+
+        melt_methods = [MeltMethodSetting.model_validate(e) for e in nut_05["methods"]]
+
+        lowest_fee: Optional[int] = None
+        for entry in sm:
+            supported = any(
+                m.method == entry.mn and m.unit == unit.name for m in melt_methods
+            )
+            if supported:
+                fee = entry.mf or 0
+                if lowest_fee is None or fee < lowest_fee:
+                    lowest_fee = fee
+
+        return lowest_fee
 
     def supports_websocket_mint_quote(self, method: Method, unit: Unit) -> bool:
         if not self.nuts or not self.supports_nut(WEBSOCKETS_NUT):
--- a/cashu/core/mint_info.py
+++ b/cashu/core/mint_info.py
@@ -3,7 +3,7 @@ from typing import Any, Dict, List, Optional
 
 from pydantic import BaseModel
 
-from .base import Method, SupportedMethod, Unit
+from .base import Method, Unit
 from .json_rpc.base import JSONRPCSubscriptionKinds
 from .models import (
     MeltMethodSetting,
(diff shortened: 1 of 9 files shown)


a1denvalu3:
@KvngMikey If I am not mistaken this only adds support for decoding such payment requests with the additional fields but does not add any logic to handle them or to create a payment request with them.

KvngMikey:
> @KvngMikey If I am not mistaken this only adds support for decoding such payment requests with the additional fields but does not add any logic to handle them or to create a payment request with them.

@a1denvalu3 you're right that this PR only adds encoding/decoding for the new ms / fr / sm fields and doesn't add enforcement logic.

Happy to either extend this PR with that handling + a request-creation path, or land this as the serialization layer and do the behavior in a focused follow-up. Which would you prefer?

a1denvalu3:
> Happy to either extend this PR with that handling + a request-creation path

Yes please

robwoodgate:
The nut has been updated: `ms` (mint_strict) is now flipped to `mp` (mint_preferred) to avoid the omitted=true antipattern.

a1denvalu3 (dismissed):

a1denvalu3:
I think there are new changes to the spec @KvngMikey

robwoodgate:
Noted by my agent while checking compliance to current spec:

- Pay flow currently never reads pr.u: add "reject if pr.sm set and pr.u absent" plus "reject if pr.u != wallet.unit"
  before calling payment_request_method_fee(pr.sm, wallet.unit). With that guard the existing wallet-unit resolution becomes correct as-is.
  - Optionally a Pydantic validator on the model (u required if a or sm); the CLI request command already always sets u, so creation side is fine in practice.

a1denvalu3:
- [P1] Reject negative method fees during validation — cashu/core/
    base.py:1681-1683
    A NUT-18 request can supply a negative mf, which passes model validation and is selected as the
    lowest supported fee. When the fee applies, pay adds this negative value to amount_to_pay,
    causing the wallet to send less than the requested amount. Constrain mf to the NUT-26 u64 range
    so both encodings share safe semantics.

  - [P2] Require a positive payment-request amount — cashu/wallet/cli/
    cli.py:529-530
    The new request command accepts zero and negative amounts, so commands such as cashu request -1
    emit requests that the corresponding pay path cannot process correctly (0 is rejected and
    negative values reach send). Use a positive integer range at argument parsing time.

  - [P2] Validate the NUT-26 mint-preferred flag length — cashu/core/nuts/
    nut26.py:342-343
    For a malformed NUT-26 request whose 0x09 field is empty, indexing val[0] raises IndexError;
    multi-byte values are also silently accepted despite the field being specified as u8. Validate
    that the field is exactly one byte, as is already done for the adjacent single-use flag.

robwoodgate:
I had my agent give this an initial look-over. The findings:

### Code review

Reviewed against [NUT-18](https://github.com/cashubtc/nuts/blob/e0a16f8/18.md) and [NUT-26](https://github.com/cashubtc/nuts/blob/e0a16f8/26.md) at current `nuts` main (`e0a16f8`), which includes the spec consolidation this implements (cashubtc/nuts#381).

Both frozen vectors check out byte-for-byte, and the `mp` strict/preferred logic, the fee applicability rule, the lowest-`mf`-among-supported-methods rule, and the `sum(proofs) - input_fee >= a + mf` invariant all match the spec. Matching `mn` against NUT-05 `method` rather than the new `method_name` is right too. Three issues:

**1. Mint URL compared by exact string equality, so a trailing-slash mismatch silently overpays**

Before this PR a mismatch produced a loud "not accepted" error. Now, with `mp=true`, the same mismatch classifies the payer's own mint as outside the list, and the payer silently pays `mf` more than they owe. The codebase already normalises for this comparison elsewhere ([`npc.py#L160-L162`](https://github.com/cashubtc/nutshell/blob/27d35dd057dedb6512e3a37ff362419c9e42065a/cashu/wallet/npc.py#L160-L162) uses `rstrip("/")`).

https://github.com/cashubtc/nutshell/blob/27d35dd057dedb6512e3a37ff362419c9e42065a/cashu/wallet/cli/cli.py#L303-L306

**2. NUT-05 `disabled` is ignored when deciding whether the mint supports a requested method**

The point of `sm` is that the receiver can melt out of the sender's mint ("the payer MUST send ecash from a mint that supports melting the request unit"). If that mint has melting switched off, that is false, but the payer still proceeds. Nutshell's own mint emits the flag next to `methods` ([`features.py#L112-L117`](https://github.com/cashubtc/nutshell/blob/27d35dd057dedb6512e3a37ff362419c9e42065a/cashu/mint/features.py#L112-L117)), so it is available here.

https://github.com/cashubtc/nutshell/blob/27d35dd057dedb6512e3a37ff362419c9e42065a/cashu/core/mint_info.py#L86-L91

**3. Fee message is wrong when the request carries no mint list**

`fee_applies` is true when `pr.m is None`, but the message then claims the payer is outside a preferred list that does not exist.

https://github.com/cashubtc/nutshell/blob/27d35dd057dedb6512e3a37ff362419c9e42065a/cashu/wallet/cli/cli.py#L335-L342

Minor: `SupportedMethod.mn` is not constrained to non-empty, so `--method ":5"` encodes an empty method name; and the `payment_request_method_fee` docstring omits the `sm` empty/None -> `0` case, which is what separates "no requirement" from `None` = "unsupported".

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cashu/core/mint_info.py
```diff
@@ -68,6 +73,38 @@ def supports_mpp(self, method: str, unit: Unit) -> bool:
...
+    def payment_request_method_fee(
+        self, sm: Optional[List[SupportedMethod]], unit: Unit
+    ) -> Optional[int]:
+        """
+        Given the `sm` (supported methods) list from a NUT-18/26 payment request,
+        return the lowest per-method fee this mint owes for melting via one of the
+        listed methods in `unit`. Returns 0 when `sm` is unset, i.e. the request
+        makes no method requirement, and `None` when this mint cannot melt via any
+        of the listed methods.
+        """
```
callebtc: does this really belong into `mint_info.py`?

KvngMikey: fair point, looking at it again, maybe not all of it should live here as it really does 2 things in 1. I'll leave the part that parses `nut[5]` and honors the `disabled` in this file and move the rest out.

PR #1070, opened 2026-06-29
fix(mint): align mint quote accounting and fix pending state priority

a1denvalu3:
This PR fixes the following:
1. **PENDING State Priority**: Ensures that if a mint quote's underlying state is `PENDING`, it is not overridden/incorrectly evaluated as `PAID`, `ISSUED`, or `UNPAID` by the dynamic state property's accounting calculation.
2. **Paid Time persistence for unpaid quotes in wallet**: Corrects the local wallet DB update where unpaid quotes would unconditionally get `paid_time` set to the current timestamp.
3. **Backwards-compatible mint quote accounting alignment**: Fields such as `amount_paid`, `amount_issued`, and `updated_at` are seamlessly synced, supported by corresponding migrations and protocol mapping, and verified by comprehensive unit tests.


callebtc:
Thanks for the update. I re-reviewed the current PR head after the wallet migration None fix. The previous migrated-wallet `None` comparison crash looks fixed now.

Remaining issues I still see:

1. **Wallet still parses `PENDING` as `PAID` when accounting fields are present**

   In `MintQuote.from_resp_wallet()`, state is still derived from accounting before honoring a response state of `PENDING`. A response like:

   ```python
   state="PENDING", amount_paid=100, amount_issued=0
   ```

   parses as `PAID` in the wallet. Mint-side gating is now protected because `MintQuote.state` prioritizes `PENDING`, but wallet display/storage and retry behavior can still be wrong. Simplest fix is to give response `PENDING` the same priority in `from_resp_wallet()` that the model property now has.

2. **`updated_at` is still not bumped on mint quote state transitions**

   The paid transition sets `state`, `paid_time`, and `last_checked`, but does not set `updated_at`. The issued/unpending paths and management RPC state update similarly update state without refreshing `updated_at`.

   Since this PR persists `updated_at`, it should be updated in the actual transition paths, not in the model setter. Otherwise clients can receive stale `updated_at` values even though the quote moved to `PAID` or `ISSUED`.

So current status from my side: P1 fixed; these two P2s remain.

KvngMikey (approved):

callebtc (approved):

PR #1071, opened 2026-06-30
fix(mint): async melt pending race

orangeshyguy21:
## Bug Fixes

- `async_melt` now runs part of the process synchronously to update the melt quote record to PENDING before returning the PENDING response. 

This prevents a race condition where a wallet receives a `PENDING` response on the melt quote `POST`, then reads a stale `UNPAID` state on a subsequent poll or subscription. This can cause wallets to revert proofs the mint later spends.

## Changes

- **Melt refactor** — split the monolithic `melt()` into `_prepare_melt()` (validation + set-pending + internal settle) and `_execute_melt_payment()` (Lightning payment + finalize); synchronous melt path behavior is unchanged.

## Testing

- **Pending-commit regression test** — asserts the persisted melt-quote state is already PENDING the moment `async_melt` returns, holding the fake backend payment in-flight to expose the window.

### DIFF
--- a/cashu/mint/ledger.py
+++ b/cashu/mint/ledger.py
@@ -1023,7 +1023,9 @@ class Ledger(
             PostMeltQuoteResponse: Melt quote response with pending state.
         """
         # validate and set the quote to pending before responding
-        melt_quote = await self._prepare_melt(proofs=proofs, quote=quote, outputs=outputs)
+        melt_quote = await self._prepare_melt(
+            proofs=proofs, quote=quote, outputs=outputs
+        )
 
         # run the payment in the background
         async def melt_task():
@@ -1058,7 +1060,9 @@ class Ledger(
         Returns:
             PostMeltQuoteResponse: Melt quote response.
         """
-        melt_quote = await self._prepare_melt(proofs=proofs, quote=quote, outputs=outputs)
+        melt_quote = await self._prepare_melt(
+            proofs=proofs, quote=quote, outputs=outputs
+        )
         return await self._execute_melt_payment(melt_quote, proofs, outputs)
 
     async def _prepare_melt(
--- a/cashu/mint/ledger.py
+++ b/cashu/mint/ledger.py
@@ -1022,24 +1022,21 @@ class Ledger(
         Returns:
             PostMeltQuoteResponse: Melt quote response with pending state.
         """
-        # get melt quote
-        melt_quote = await self.get_melt_quote(quote_id=quote)
-        if not melt_quote:
-            raise TransactionError("melt quote not found")
-        if not melt_quote.unpaid:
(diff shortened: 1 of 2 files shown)


cashu/mint/ledger.py
```diff
@@ -1022,24 +1022,23 @@ async def async_melt(
...
             except Exception as e:
                 logger.error(f"Error in background melt task: {e}")
 
         asyncio.create_task(melt_task())
 
-        melt_quote.state = MeltQuoteState.pending
-        return PostMeltQuoteResponse.from_melt_quote(melt_quote)
+        # respond pending now that it is durably committed
+        pending_quote = melt_quote.model_copy(update={"state": MeltQuoteState.pending})
+        return PostMeltQuoteResponse.from_melt_quote(pending_quote)
```
callebtc, line 1041: is it necessary to set the melt quote pending in async_melt if prepare_melt does it too?

orangeshyguy21: Yes it's needed as written, but only for the internal-settlement path. The two paths through
`async_melt`:

    # Lightning path (external invoice)
    _prepare_melt()           → quote committed PENDING
    respond PENDING           → benign
    background task           → pay invoice, then change + invalidate proofs → PAID

    # internal-settlement path (invoice belongs to this mint)
    _prepare_melt()           → quote committed PAID (melt_mint_settle_internally)
    respond PENDING           → this is where the line matters
    background task           → skips payment, still does change + invalidate proofs

On the internal path the quote is PAID but the melt isn't finished. Change promises
and proof invalidation still happen in the background. Always responding PENDING from `async_melt` keeps the same contract as before.

It's a `model_copy` because the background task holds the same object and
`_execute_melt_payment` branches on `melt_quote.paid`. Flipping PAID→PENDING on the object
could make it try to pay an already-settled quote using the backend.

At a minimum, code comments should be improved here.

There is an alternative refactor where this double set PENDING logic could be avoided. Don't background anything if we are settling the melt internally:

```
melt_quote = await self._prepare_melt(proofs=proofs, quote=quote, outputs=outputs)

if melt_quote.paid:
    # settled internally — nothing async left to do; finish now and
    # return the real PAID response, change included
    return await self._execute_melt_payment(melt_quote, proofs, outputs)

# real Lightning payment → background it, respond PENDING
asyncio.create_task(melt_task())
return PostMeltQuoteResponse.from_melt_quote(melt_quote)  # already PENDING, no copy needed
```

callebtc:
Superseded by #1087 (same race, cleaner cut).

Prefer #1087: `_prepare_melt` only locks durable `PENDING` + blanks; internal settle stays in `_execute_melt_payment`, so the async response matches DB state without forcing PENDING via `model_copy` when internal settle already marked the quote PAID.

PR #1075, opened 2026-07-03
chore: migrate dependency management to uv

callebtc:
## Summary
- Migrate packaging and dependency management from Poetry to uv with PEP 621 metadata.
- Update CI, Docker, Makefile, docs, and helper scripts to use uv.
- Replace poetry.lock with uv.lock and regenerate requirements.txt from uv.

## Verification
- uv sync --locked --python 3.10.4
- uv run python scripts/check_version.py v0.20.2
- uv run cashu --help; uv run mint --help; uv run mint-cli --help
- make ruff-check
- uv build --quiet
- pytest tests/fuzz/test_fuzz_mint_api.py tests/test_core.py tests/test_crypto.py -q
- pytest tests/mint/test_mint_app_router.py tests/mint/test_mint_limit.py tests/mint/test_mint_management_rpc_server.py -q

Full make test was also attempted locally; the broader suite did not complete cleanly because SQLite fixture disk I/O errors appeared after many passing tests.


a1denvalu3 (commented):
1. Remove "setuptools" and "wheel" from standard runtime dependencies in pyproject.toml as they are only required during the build step and are already specified in [build-system].
2. Run `uv export --format requirements.txt --no-dev --no-hashes --no-emit-project --output-file requirements.txt`

PR #1076, opened 2026-07-03
chore: update dependencies

callebtc:
## Summary
- update advisory-related Python dependencies to current compatible releases
- adapt wallet HTTPX proxy setup and template rendering for newer APIs
- harden live mint test fixture around readiness, SQLite reset, and Redis cache isolation

## Verification
- pytest tests -q -x --tb=short
- make check
- standalone mint smoke against /v1/info


requirements.txt
vicjuma, line 1: I'm wondering whether we could simplify this by just reconciling pyproject.toml ↔ requirements.txt using the Makefile's **`package`** target

```makefile
package:
	poetry export -f requirements.txt --without-hashes --output requirements.txt
	make clean
	python setup.py sdist bdist_wheel
```
Or maybe just create another target if we only want a minimal export command

```makefile
requirements:
	poetry export -f requirements.txt --without-hashes --output requirements.txt

package: requirements
	$(MAKE) clean
	python setup.py sdist bdist_wheel

.PHONY: requirements package
```
I may be misunderstanding the intent behind this manual update — could you clarify? IMO this shouldn't be committed at all if it can be generated during the setup/packaging process.

PR #1081, opened 2026-07-14
feat(wallet): add efficient recovery

a1denvalu3:
## Summary

- implement efficient wallet recovery proposed in [cashubtc/nuts#342](https://github.com/cashubtc/nuts/pull/342)
- add encrypted `d_gap` metadata to blinded messages and signatures
- persist and restore `d_gap` at the mint and advertise NUT-342 support
- calculate recovery gaps from unspent and pending proofs within the same keyset
- recover using windowed binary-search probes with NUT-13 fallback

## Implementation notes

Binary-search probes use the legacy recovery batch size as their window size. This tolerates isolated derivation gaps while keeping the strategy wallet-defined and wire-compatible.

The wallet reuses the derivation counter already stored in each proof's `derivation_path`; no wallet schema migration is needed. The mint adds a nullable `d_gap` column to `promises`.

## Testing

- `poetry run ruff check .`
- `poetry run mypy cashu --check-untyped-defs`
- focused efficient-recovery, wallet restore, mint database, and auth tests

## Review considerations

- draft specification is still under discussion
- includes a mint database migration
- includes AES-128-GCM recovery metadata keyed from the deterministic blinding factor
- windowed binary search tolerates gaps within a probe window; wallets fall back to NUT-13 when the search assumptions cannot be established


a1denvalu3:
image

PR #1085, opened 2026-07-16
refactor(mint): recalculate DLEQs on restore

a1denvalu3:
## Summary

- stop persisting deterministic DLEQ proof values with mint promises
- drop the `dleq_e` and `dleq_s` columns from mint and auth-mint databases
- regenerate and validate DLEQ proofs when a client calls the restore endpoint
- verify restored deterministic proofs match the originally issued proofs

## Rationale

DLEQ nonces are now derived deterministically from the mint private key and proof context. The mint can therefore reconstruct the proof from the stored blinded signature metadata and keyset instead of retaining `e` and `s` in the database.

Promises issued before deterministic nonce generation may receive a different DLEQ after restore. This remains valid: DLEQ proof serialization is not an identity for the blind signature, and clients verify the returned proof against the unchanged `A`, `B'`, and `C'` values.

## Review notes

This includes database migrations and cryptographic proof handling, so it requires maintainer review under the repository contribution guidance. The migration only removes derived proof data; it does not alter stored blind signatures or key material.

## Testing

- `poetry run pytest tests/mint/test_mint_migrations.py tests/mint/test_mint_api.py::test_api_restore -q` — 3 passed
- `make check` — Ruff and mypy passed

### DIFF
--- a/cashu/mint/crud.py
+++ b/cashu/mint/crud.py
@@ -183,8 +183,6 @@ class LedgerCrud(ABC):
         amount: int,
         b_: str,
         c_: str,
-        e: str = "",
-        s: str = "",
         conn: Optional[Connection] = None,
     ) -> None: ...
 
@@ -443,15 +441,13 @@ class LedgerCrudSqlite(LedgerCrud):
         await (conn or db).execute(
             f"""
             UPDATE {db.table_with_schema("promises")}
-            SET amount = :amount, c_ = :c_, dleq_e = :dleq_e, dleq_s = :dleq_s, signed_at = :signed_at
+            SET amount = :amount, c_ = :c_, signed_at = :signed_at
             WHERE b_ = :b_;
             """,
             {
                 "b_": b_,
                 "amount": amount,
                 "c_": c_,
-                "dleq_e": e,
-                "dleq_s": s,
                 "signed_at": db.to_timestamp(db.timestamp_now_str()),
             },
         )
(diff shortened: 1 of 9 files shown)


cashu/mint/ledger.py
```diff
@@ -1340,13 +1340,23 @@ async def restore(
...
                 if promise is not None:
+                    promise.dleq = self._generate_dleq(output, promise)
                     signatures.append(promise)
                     return_outputs.append(output)
                     logger.trace(f"promise found: {promise}")
         return return_outputs, signatures
 
     # ------- BLIND SIGNATURES -------
 
+    def _generate_dleq(self, output: BlindedMessage, promise: BlindedSignature) -> DLEQ:
```
vicjuma, line 1360: Tested wallet restore acknowledged. I like this refactor. It removes unnecessary parameter threading of derived state across multiple layers just to persist it. `from_row()` is cleaner as well, leveraging the optional dleq field while keeping the database mapping focused on canonical data.

```txt
pytest tests/wallet/test_wallet_restore.py::test_restore_wallet_after_mint
```

cashu/mint/crud.py
KvngMikey: the method's signature should drop `e` and `s` here also, with any test updated to reflect this also.

cashu/mint/ledger.py
```diff
@@ -1340,13 +1349,23 @@ async def restore(
...
```
KvngMikey, line 1360: if a promise references a keyset or amount not in memory we get a bare `KeyError (500)`, but `ledger._sign_blinded_messages` has an explicit "keyset not found" TransactionError for the same lookup, i think _generate_dleq can match that.

KvngMikey (approved):

KvngMikey:
ACK, LGTM!

PR #1088, opened 2026-07-17
feat: verify and sign the length-framed NUT-11 SIG_ALL message

robwoodgate:
Implements the length-framed NUT-11 SIG_ALL message aggregation (`Cashu_SigAllSig_v1`) proposed in https://github.com/cashubtc/nuts/pull/404. The framing matches NUT-20 mint quote signatures; the message is signed over its BIP-340 tagged hash, the same construction NUT-06 signed mint info (cashubtc/nuts#416) uses.

- The mint verifies the v1 and current message formats as digests (the v1 tagged hash and the SHA-256 of the current string), counts unique public keys with valid signatures, and ignores signatures that do not verify (including malformed ones, which previously aborted the verification loop)
- The wallet signs the v1 and current formats, so its transactions verify on mints on either side of the upgrade
- Melt verification passes the quote id instead of a prebuilt message string; each format binds the quote per its own rules
- Canonical vectors from the spec's `tests/11-test.md` are pinned in `tests/nuts/test_nut11_test_vectors.py`


KvngMikey (approved):
ACK @ https://github.com/cashubtc/nutshell/pull/1088/commits/011159499d58d7119eec0c7f5cc00f2ae726747c

Just the final agreement on the legacy format left.

robwoodgate:
> ACK @ [0111594](https://github.com/cashubtc/nutshell/commit/011159499d58d7119eec0c7f5cc00f2ae726747c)
> 
> Just the final agreement on the legacy format left.

My recommendation would be current ('v0') and the proposed 'v1' only, the legacy MTS should be dropped unilaterally. SIG_ALL is a pretty niche feature, so I think the benefit of dropping it outweighs the inconvenience to older wallets.

robwoodgate:
I have removed the legacy signing format this PR had re-added

a1denvalu3 (approved):

cashu/wallet/p2pk.py
```diff
@@ -168,24 +169,29 @@ def add_witness_swap_sig_all(
...
-            message_to_sign = message_to_sign or nut11.sigall_message_to_sign(
-                proofs, outputs
-            )
-            # For P2BK proofs, use the derived blinded signing key
+            # Sign every accepted SIG_ALL message format so the transaction
+            # verifies on mints that have not (or have already) upgraded. Mints
+            # ignore signatures that do not verify and count unique pubkeys.
+            quote_suffix = quote_id or ""
+            messages_to_sign: List[Union[str, bytes]] = [
+                nut11.sigall_message_to_sign_v1(proofs, outputs, quote_id),
```
a1denvalu3, line 177: what exactly is v1? should this be called legacy? legacy code should be clearly wrapped in a BEGIN ... END legacy section

robwoodgate: "v1" is the new version specified (`Cashu_SigAllSig_v1`), length framed MTS, see `cashu/core/nuts/nut11.py#L13`

The legacy MTS was removed, the outgoing (currently live, unversioned, not length prefixed) MTS as below, now at line 35 of nut11.py:

```
nut11.sigall_message_to_sign(proofs, outputs) + quote_suffix,
```

robwoodgate:
UPDATED: https://github.com/cashubtc/nutshell/pull/1088/changes/9dbb884b78e968af93c63d2792746deef7149b43

This commit brings Nutshell in line with latest revision to spec, which is replacing the flat DST with a tagged hash

Regtest suite seems to be private and unavailable to the CI runners, which is why they timed out

PR #1090, opened 2026-07-18
feat(cli): add --timelock/-t option to send command

KvngMikey:
## Summary

`send`'s locktime was env-var-only: `settings.locktime_delta_seconds` was passed unconditionally to `create_p2pk_lock`/`create_p2bk_lock` with no per-invocation way to override it. Testing a short refund window required setting `LOCKTIME_DELTA_SECONDS=10` as an environment variable at send time.

The inconsistency this resolves: `cashu lock p2pk` already exposes `--timelock/-t` and prints a `cashu send <amount> --lock <lock>` command as next-step guidance, but that guidance silently dropped the user's `--timelock` value, because `send` had no flag to carry it.

**Fix**: add `--timelock/-t <seconds>` to `send`, threaded through `helpers.send()` as an optional override (`Optional[int] = None`) that falls back to `settings.locktime_delta_seconds` when omitted. Also fixed `lock_p2pk`'s printed `send` command to include `--timelock <n>` when the user set one.

- This UX feature was found while interop-testing NUT-28 (https://github.com/cashubtc/nutshell/pull/950);
- Tests added to confirm fix works
- Also tested manually after the fix confirming it works and no existing feature was broken.

### DIFF
--- a/cashu/wallet/helpers.py
+++ b/cashu/wallet/helpers.py
@@ -157,7 +157,7 @@ async def send(
             if timelock_seconds is not None
             else settings.locktime_delta_seconds
         )
-        # we add a time lock to the P2PK lock by appending the current unix time + 14 days
+        # we add a time lock by appending the current unix time + locktime_seconds
         if lock.startswith("P2PK:") or lock.startswith("P2PK-SIGALL:"):
             sigall = lock.startswith("P2PK-SIGALL:")
             logger.debug(f"Locking token to: {lock}")
--- a/cashu/wallet/helpers.py
+++ b/cashu/wallet/helpers.py
@@ -141,6 +141,7 @@ async def send(
     memo: Optional[str] = None,
     force_swap: bool = False,
     refund_pubkeys: Optional[List[str]] = None,
+    timelock_seconds: Optional[int] = None,
 ):
     """
     Prints token to send to stdout.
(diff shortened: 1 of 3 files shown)


cashu/wallet/helpers.py
```diff
@@ -151,20 +152,23 @@ async def send(
         assert len(lock) > 21, Exception(
             "Error: lock has to be at least 22 characters long."
         )
```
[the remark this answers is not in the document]
KvngMikey: out of scope for this PR, that assert predates this change and isn't touched by it. Happy to revisit in a separate cleanup if wanted.

a1denvalu3 (approved):

cashu/wallet/helpers.py
```diff
@@ -173,16 +177,14 @@ async def send(
...
-            )
+            logger.debug(f"Adding a time lock of {locktime_seconds} seconds.")
             tags = None
             if refund_pubkeys:
                 tags = Tags()
                 tags["refund"] = refund_pubkeys
             secret_lock, p2pk_e = await wallet.create_p2bk_lock(
                 lock.split(":")[1],
-                locktime_seconds=settings.locktime_delta_seconds,
+                locktime_seconds=locktime_seconds,
```
a1denvalu3, line 187: Does this **always** set a locktime, even if we didn't pass any?

KvngMikey: yes, and that's pre-existing behavior of default locktime to 1 day, that's what makes the refund path possible at all.

this PR only changes which value is threaded in, the --timelock override when given, otherwise exactly the same settings.locktime_delta_seconds fallback.

PR #1095, opened 2026-07-23
feat(lightning): probe LND routing fees

a1denvalu3:
## Summary

- query LND’s `EstimateRouteFee` API when creating full-invoice melt quotes
- support both the REST and gRPC LND backends
- use the probed route fee with the configured minimum reserve as a floor
- fall back to the existing percentage/minimum reserve when probing fails
- keep partial-MPP quotes on the existing reserve because LND’s invoice probe estimates the full invoice
- add real multi-hop regtest coverage usable with both LND transports

## Testing

- `make check`
- `pytest -q tests/lightning/test_lightning_backends_mocked.py tests/lightning/test_fee_unit_regressions.py` (37 passed)
- real cashu-regtest LND 0.21 REST: `test_lnd_payment_quote_uses_probed_fee` (passed)
- real cashu-regtest LND 0.21 gRPC: `test_lnd_payment_quote_uses_probed_fee` (passed)

The regtest requests a 100,000 sat CLN node 2 invoice that requires a multi-hop route from LND node 3. It asserts that the returned fee is non-zero and lower than the blind percentage-based reserve.

### DIFF
--- a/cashu/lightning/lnd_grpc/lnd_grpc.py
+++ b/cashu/lightning/lnd_grpc/lnd_grpc.py
@@ -48,6 +48,7 @@ INVOICE_RESULT_MAP = {
 
 MAX_ROUTE_RETRIES = 50
 PAYMENT_TIMEOUT_SECONDS = 60
+FEE_PROBE_TIMEOUT_SECONDS = 5
 
 
 class LndRPCWallet(LightningBackend):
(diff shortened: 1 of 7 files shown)


a1denvalu3:
Rationale for the updated fee handling:

`EstimateRouteFee` explicitly returns a lower bound, so using the probe result by itself as the melt fee reserve could underquote when LND ultimately selects a different or more expensive route. The quote now adds `lightning_reserve_fee_min` to the probed fee, using the configured base reserve as a buffer for route changes while still producing a substantially tighter quote than the blind percentage-based reserve.

At payment time, the mint already converts the stored quote reserve to msat and passes that exact value to LND as `fee_limit_msat`. This makes the user-visible quote an enforced upper limit: LND rejects routes above it before settlement. The regtest now pays the probed multi-hop invoice and verifies that the actual settled fee is less than or equal to the quoted limit. If probing fails or times out, behavior still falls back to the existing configured percentage-based reserve.

callebtc (commented):
very nice. not ready yet?

PR #1096, opened 2026-07-24
fix: check final_expiry != 0

asmogo:


asmogo:
We could instead check for `expiry > 0`, which aligns with the CDK implementation. https://github.com/cashubtc/cdk/blob/main/crates/cashu/src/nuts/nut02.rs#L206

a1denvalu3:
This is not retro-compatible and would cause issued with keysets that were generated previously.

PR #1105, opened 2026-08-04
feat(mint): support BOLT11 description hashes

sotach1:
## Summary

- add an optional per-quote `description_hash` flag to the NUT-23 BOLT11 mint quote request
- advertise backend support as `options.description_hash` in mint info
- preserve ordinary plaintext invoice descriptions unless the wallet explicitly opts in
- pass both the computed hash and original description bytes to capable Lightning backends
- verify that the backend returned an invoice with the requested description hash

This supports protocols such as NIP-57 where the invoice must commit to exact data supplied out of band. The behavior is deliberately per quote: ordinary Cashu deposit descriptions continue to use the BOLT11 `d` tag, while an opted-in quote uses the `h` tag.

Protocol proposal: cashubtc/nuts#417

The capability is currently enabled for FakeWallet, LND gRPC/REST, LNbits, and Blink. Other backends reject opted-in requests rather than returning an invoice with the wrong commitment.

## Request example

```json
{
  "amount": 21,
  "unit": "sat",
  "description": "<exact out-of-band data>",
  "description_hash": true
}
```

## Testing

- Ruff: all checks passed
- mypy: 131 source files passed
- targeted mint description tests: 3 passed (plaintext, hashed, missing-description rejection)
- lnuts integration suite: 21 passed, including a NIP-57 client using the per-quote flag
- live LND plaintext check: invoice retained its exact `d` description and had no `h` tag
- live LND hashed check: invoice had no `d` tag and its `h` tag matched the exact description
- live NIP-57 check: `so@nuts.cash` returned a 42-sat invoice whose `h` tag matched the exact signed kind-9734 JSON

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -142,6 +142,8 @@ MINT_RPC_SERVER_CA="./ca_cert.pem"
 # MINT_BOLT11_DISABLE_MINT=FALSE
 # Disable melting of BOLT11 invoices
 # MINT_BOLT11_DISABLE_MELT=FALSE
+# Encode NUT-23 descriptions as BOLT11 description hashes instead of plaintext.
+# MINT_BOLT11_HASH_DESCRIPTIONS=FALSE
 
 # Rate limit requests to mint (enabled by default).
 # Set MINT_RATE_LIMIT=FALSE to disable.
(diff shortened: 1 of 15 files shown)


KvngMikey:
@sotach1 a couple failed tests you should look at.

PR #1114, opened 2026-08-12
feat(mint): loosen NUT-29 batch quote checks

a1denvalu3:
## Summary

- return exactly one NUT-29 batch-check entry per requested quote ID, preserving request order
- return positional `{ "quote": <id>, "unknown": true }` entries for unknown or malformed quote IDs
- use the NUT-04 accounting response fields and omit the deprecated `state` field from batch-check quote responses
- validate batch-mint eligibility using the currently mintable amount (`amount_paid - amount_issued`)
- add ledger and HTTP API coverage for mixed known/unknown and all-unknown requests

Implements the proposed changes in cashubtc/nuts#422 at spec commit `ce8539c7`. This is protocol-sensitive and remains a draft pending maintainer review.

## Validation

- `MINT_BACKEND_BOLT11_SAT=FakeWallet TOR=FALSE poetry run pytest tests/mint/test_mint_batch.py tests/mint/test_mint_api.py -q` (40 passed, 2 skipped)
- latest behavior-focused tests (5 passed)
- `poetry run ruff check cashu/core/models/mint_quote.py cashu/core/models/__init__.py cashu/mint/router.py cashu/mint/ledger.py tests/mint/test_mint_batch.py tests/mint/test_mint_api.py`
- `poetry run mypy cashu/core/models/mint_quote.py cashu/mint/router.py cashu/mint/ledger.py`
- `git diff --check`

### DIFF
--- a/cashu/core/models/__init__.py
+++ b/cashu/core/models/__init__.py
@@ -30,6 +30,8 @@ from .mint import (
 )
 from .mint_quote import (
     PostMintQuoteCheckRequest,
+    PostMintQuoteCheckResponse,
+    PostMintQuoteCheckUnknownResponse,
     PostMintQuoteRequest,
     PostMintQuoteResponse,
 )
@@ -62,6 +64,8 @@ __all__ = [
     "PostMintRequest",
     "PostMintResponse",
     "PostMintQuoteCheckRequest",
+    "PostMintQuoteCheckResponse",
+    "PostMintQuoteCheckUnknownResponse",
     "PostMintQuoteRequest",
     "PostMintQuoteResponse",
     "PostRestoreRequest",
--- a/cashu/core/models/mint_quote.py
+++ b/cashu/core/models/mint_quote.py
@@ -1,4 +1,4 @@
-from typing import Annotated, List, Optional
+from typing import Annotated, List, Literal, Optional
 
 from pydantic import BaseModel, Field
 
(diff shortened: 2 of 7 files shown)


a1denvalu3:
Security follow-up: the double-mint races from cashubtc/security-audits#392 were reproduced against the rebased head (96 sats issued for a 64-sat payment, both PoC variants) and fixed in 1b388f69.

1. **TOCTOU guard** — `_set_mint_quote_pending()` now rejects partially-issued quotes under the row lock; a concurrent batch mint can no longer race the stale-read guard in `mint()`.
2. **Accounting-clobbering revert/finalize** — the singular `_unset_mint_quote_pending()` bypasses the `MintQuote.state` setter (no more `amount_issued` reset to 0 on revert), and `mint()` finalizes via `_issue_mint_quotes()`, which increments `amount_issued` atomically instead of overwriting it.
3. **Stale batch validation** — `_issue_mint_quotes()` re-validates each amount against the currently mintable amount under the row lock; a failed issuance now reverts quotes to PAID (accounting intact) instead of leaving them stuck in PENDING.

Verification: both PoCs now fail at the hardened lines; three regression tests added in `tests/mint/test_mint_batch.py`; full mint suite passes (568 passed, 43 skipped, FakeWallet); ruff/mypy clean. The optional setter refactor (point 4) was left out — other paths rely on it.

PR #1122, opened 2026-08-15
Allow minting paid quotes after expiry

Egge21M:
## Summary

- remove the issuance-time expiry rejection from single and batch minting
- keep payment-state, amount, witness, pending, and issued checks unchanged
- add regression coverage for minting paid quotes after their expiry timestamp

## Root cause

Nutshell records a mint quote as paid when its Lightning backend reports the invoice settled, but `mint` and `mint_batch` independently rejected that paid quote once its expiry timestamp passed. This could leave a successfully paid quote permanently unmintable if the wallet did not request signatures before expiry.

Quote expiry limits when the payment request can be paid. Once payment is confirmed, the paid value should remain mintable until it has been issued.

## Impact

Wallets can recover ecash from paid quotes after expiry, including through the batch endpoint. Unpaid quotes remain unmintable, and the existing state transitions continue to prevent duplicate issuance.

## Validation

- `pytest tests/mint/test_mint.py tests/mint/test_mint_batch.py -q` — 37 passed
- `ruff check cashu/mint/ledger.py tests/mint/test_mint.py tests/mint/test_mint_batch.py`

### DIFF
--- a/cashu/mint/ledger.py
+++ b/cashu/mint/ledger.py
@@ -496,7 +496,6 @@ class Ledger(
             Exception: Validation of outputs failed.
             Exception: Quote not paid.
             Exception: Quote already issued.
-            Exception: Quote expired.
             Exception: Amount to mint does not match quote amount.
 
         Returns:
@@ -515,6 +514,9 @@ class Ledger(
         if quote.state != MintQuoteState.paid:
             raise QuoteNotPaidError()
 
+        # Quote expiry limits when the payment request can be paid. Once payment is
+        # confirmed, the quote remains mintable until its value has been issued.
+
         previous_state = quote.state
         await self.db_write._set_mint_quote_pending(quote_id=quote_id)
         try:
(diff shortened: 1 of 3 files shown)


cashu/mint/ledger.py
```diff
@@ -592,15 +592,17 @@ async def mint_batch(
...
-            if quote.state != MintQuoteState.paid:
-                raise QuoteNotPaidError()
-
-        # Check amount balance
+            if quote.state != MintQuoteState.paid:
+                raise QuoteNotPaidError()
+
+        # Quote expiry limits payment, not issuance of already-paid value.
+
+        # Check amount balance
```
callebtc, line 605: this change isn't needed

cashu/mint/ledger.py
```diff
@@ -512,19 +511,20 @@ async def mint(
...
-            raise QuoteNotPaidError()
-
-        previous_state = quote.state
+        if quote.state != MintQuoteState.paid:
+            raise QuoteNotPaidError()
+
+        # Quote expiry limits when the payment request can be paid. Once payment is
+        # confirmed, the quote remains mintable until its value has been issued.
+
+        previous_state = quote.state
```
callebtc, line 520: this change isn't needed

PR #1128, opened 2026-08-20
fix(mint): return stored signatures on idempotent POST /mint retry

### LINKED ISSUE #257: Mint should reissue tokens for a failed `POST /mint` request.
Mint should remember which outputs where signed for which invoice and in case a wallet asks repeats the requests with the same outputs, should simply reissue the tokens. That way, we can avoid scary scenarios where the connection drops during the `/mint` process. Related to: https://github.com/cashubtc/cashu-ts/issues/78

Hardeezah:
When a wallet retries `POST /mint` with the same outputs after a connection drop, the mint now returns the previously stored signatures instead of raising `OutputsAlreadySignedError`. This prevents token loss when the connection is lost during the minting process.

- Add `get_blind_signatures_mint_id()` and `get_blinded_Bs_mint_id()` to `crud.py`
- Add idempotency block in `ledger.mint()` that checks quote state first
- Verify `B_` values match (not just output count) to prevent mismatched returns
- Update `test_mint_internal` to expect success on retry
- Add tests for idempotent retry and different-output rejection

Closes #257

### DIFF
--- a/cashu/mint/crud.py
+++ b/cashu/mint/crud.py
@@ -402,6 +402,41 @@ class LedgerCrudSqlite(LedgerCrud):
         )
         return [BlindedSignature.from_row(r) for r in rows] if rows else []  # type: ignore
 
+    async def get_blind_signatures_mint_id(
+        self,
+        *,
+        db: Database,
+        mint_id: str,
+        conn: Optional[Connection] = None,
+    ) -> List[BlindedSignature]:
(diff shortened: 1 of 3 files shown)


Hardeezah:
Hi @a1denvalu3, could you review this?

ye0man:
@Hardeezah please rebase :)

PR #1129, opened 2026-08-20
Fix: Standardize error codes

### LINKED ISSUE #264: Standardize errors
Sit down, go through every single error we have so far and give them each an error code. Then return a standard error format that will be formalized in a NUT. Then go through the LNbits extension and make sure that the response is the same as the standalone mint.

### LINKED ISSUE #769: Update authentication error codes to match NUT-21/NUT-22 specifications
Authentication error codes in "cashu/core/errors.py" are using outdated codes that don't match the official NUT-21 and NUT-22 specifications.

-ClearAuthRequiredError.code = 80001
-ClearAuthFailedError.code = 80002
-BlindAuthRequiredError.code = 81001
-BlindAuthFailedError.code = 81002
-BlindAuthAmountExceededError.code = 81003
-BlindAuthRateLimitExceededError.code = 81004

Expected Behavior

According to https://github.com/cashubtc/nuts/blob/main/21.md and https://github.com/cashubtc/nuts/blob/main/22.md specifications:

Should be updated to match NUT specifications
-ClearAuthRequiredError.code = 30001      # NUT-21
-ClearAuthFailedError.code = 30002        # NUT-21
-BlindAuthRequiredError.code = 31001      # NUT-22
-BlindAuthFailedError.code = 31002        # NUT-22
-BlindAuthAmountExceededError.code = 31003 # NUT-22
-BlindAuthRateLimitExceededError.code = 31004 # NUT-22

- NUT-21 Clear Authentication: https://github.com/cashubtc/nuts/blob/main/21.md#error-codes
- NUT-22 Blind Authentication: https://github.com/cashubtc/nuts/blob/main/22.md#error-codes
- Official Error Codes Table: https://github.com/cashubtc/nuts/blob/main/error_codes.md

Files to Update

- cashu/core/errors.py - Update error codes for authentication classes

Proposed Fix

class ClearAuthRequiredError(CashuError):
      detail = "Endpoint requires clear auth"
      code = 30001  # Updated from 80001

class ClearAuthFailedError(CashuError):
      detail = "Clear authentication failed"
      code = 30002  # Updated from 80002

class BlindAuthRequiredError(CashuError):
      detail = "Endpoint requires blind auth"
      code = 31001  # Updated from 81001

class BlindAuthFailedError(CashuError):
      detail = "Blind authentication failed"
      code = 31002  # Updated from 81002

class BlindAuthAmountExceededError(CashuError):
      detail = "Maximum BAT mint amount exceeded"
      code = 31003  # Updated from 81003

class BlindAuthRateLimitExceededError(CashuError):
      detail = "BAT mint rate limit exceeded"
      code = 31004  # Updated from 81004

KvngMikey:
## Summary
- Brings `cashu/core/errors.py` and its mint-side call sites into agreement with [`cashubtc/nuts/error_codes.md`] (https://github.com/cashubtc/nuts/blob/main/error_codes.md).

- Every change is codes-only — conditions and error messages are untouched, so callers matching on type or `detail` are unaffected.

### DIFF
--- a/cashu/core/errors.py
+++ b/cashu/core/errors.py
@@ -234,7 +234,7 @@ class QuoteRequiresPubkeyError(CashuError):
 
 class ClearAuthRequiredError(CashuError):
     detail = "Endpoint requires clear auth"
-    code = 80001
+    code = 30001
 
     def __init__(self):
         super().__init__(self.detail, code=self.code)
@@ -242,7 +242,7 @@ class ClearAuthRequiredError(CashuError):
 
 class ClearAuthFailedError(CashuError):
     detail = "Clear authentication failed"
-    code = 80002
+    code = 30002
 
     def __init__(self):
         super().__init__(self.detail, code=self.code)
@@ -250,7 +250,7 @@ class ClearAuthFailedError(CashuError):
 
 class BlindAuthRequiredError(CashuError):
     detail = "Endpoint requires blind auth"
-    code = 81001
+    code = 31001
 
     def __init__(self):
         super().__init__(self.detail, code=self.code)
@@ -258,7 +258,7 @@ class BlindAuthRequiredError(CashuError):
 
 class BlindAuthFailedError(CashuError):
     detail = "Blind authentication failed"
-    code = 81002
+    code = 31002
 
     def __init__(self):
         super().__init__(self.detail, code=self.code)
@@ -266,7 +266,7 @@ class BlindAuthFailedError(CashuError):
 
 class BlindAuthAmountExceededError(CashuError):
     detail = "Maximum blind auth amount exceeded"
-    code = 81003
+    code = 31003
 
     def __init__(self, detail: Optional[str] = None):
         super().__init__(detail or self.detail, code=self.code)
@@ -274,7 +274,7 @@ class BlindAuthAmountExceededError(CashuError):
 
 class BlindAuthRateLimitExceededError(CashuError):
     detail = "Blind auth token mint rate limit exceeded"
-    code = 81004
+    code = 31004
 
     def __init__(self):
         super().__init__(self.detail, code=self.code)
--- a/cashu/core/errors.py
+++ b/cashu/core/errors.py
@@ -213,16 +213,16 @@ class QuoteNotPaidError(CashuError):
     detail = "quote not paid"
     code = 20001
 
-    def __init__(self):
-        super().__init__(self.detail, code=self.code)
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
 
 
 class QuoteAlreadyIssuedError(CashuError):
     detail = "quote already issued"
     code = 20002
 
-    def __init__(self):
-        super().__init__(self.detail, code=self.code)
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
 
 
 class MintingDisabledError(CashuError):
--- a/cashu/core/errors.py
+++ b/cashu/core/errors.py
@@ -201,9 +201,41 @@ class QuoteAlreadyIssuedError(CashuError):
         super().__init__(self.detail, code=self.code)
 
 
+class MintingDisabledError(CashuError):
+    detail = "Minting is disabled"
+    code = 20003
+
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
+
+
 class LightningPaymentFailedError(CashuError):
     detail = "Lightning payment failed"
     code = 20004
+
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
+
+
+class QuotePendingError(CashuError):
+    detail = "quote is pending"
+    code = 20005
+
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
+
+
+class InvoiceAlreadyPaidError(CashuError):
+    detail = "invoice already paid"
+    code = 20006
+
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
+
+
+class QuoteExpiredError(CashuError):
+    detail = "quote is expired"
+    code = 20007
 
     def __init__(self, detail: Optional[str] = None):
         super().__init__(detail or self.detail, code=self.code)
--- a/cashu/core/errors.py
+++ b/cashu/core/errors.py
@@ -135,6 +135,30 @@ class TransactionUnitMismatchError(TransactionError):
         super().__init__(detail, code=self.code)
 
 
+class AmountlessInvoiceNotSupportedError(TransactionError):
+    detail = "Amountless invoice is not supported"
+    code = 11011
+
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
+
+
+class AmountMismatchError(TransactionError):
+    detail = "Amount in request does not equal invoice"
+    code = 11012
+
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
+
+
+class UnitNotSupportedError(TransactionError):
+    detail = "Unit in request is not supported"
+    code = 11013
+
+    def __init__(self, detail: Optional[str] = None):
+        super().__init__(detail or self.detail, code=self.code)
+
+
 class BatchDuplicateQuotesError(TransactionError):
     detail = "Duplicate quote IDs provided"
     code = 11016
--- a/cashu/core/errors.py
+++ b/cashu/core/errors.py
@@ -75,14 +75,14 @@ class TransactionNotBalancedError(TransactionError):
 
 
 class SecretTooLongError(TransactionError):
-    code = 11003
+    code = 11000
 
     def __init__(self, detail="secret too long"):
         super().__init__(detail, code=self.code)
 
 
 class WitnessTooLongError(TransactionError):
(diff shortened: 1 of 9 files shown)


cashu/core/errors.py
```diff
@@ -75,14 +75,14 @@ def __init__(self, detail):
 
 
 class SecretTooLongError(TransactionError):
-    code = 11003
+    code = 11000
 
     def __init__(self, detail="secret too long"):
         super().__init__(detail, code=self.code)
```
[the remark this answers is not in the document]
KvngMikey: deliberate generic fall back.

a1denvalu3 (approved):

callebtc (commented):
Review findings:

- High: A repeat melt request reaches Ledger._prepare_melt before the new locked guard. Its `if not melt_quote.unpaid` returns generic 11000 for both PAID and PENDING, rather than the documented 20006 and 20005. Please split that pre-check too.

The two inline comments cover the remaining issued-quote paths.

cashu/mint/db/write.py
```diff
@@ -171,9 +175,9 @@ async def _set_mint_quote_pending(self, quote_id: str) -> MintQuote:
             if not quote:
                 raise TransactionError("Mint quote not found.")
             if quote.pending:
-                raise TransactionError("Mint quote already pending.")
+                raise QuotePendingError("Mint quote already pending.")
             if not quote.paid:
-                raise TransactionError("Mint quote is not paid yet.")
+                raise QuoteNotPaidError("Mint quote is not paid yet.")
```
callebtc, line 182: An issued quote is neither pending nor paid, so a state transition from paid to issued between the optimistic check in Ledger.mint and this lock falls through to QuoteNotPaidError (20001). The batch version below handles this correctly. Add an `if quote.issued: raise QuoteAlreadyIssuedError(...)` branch before `if not quote.paid` so the authoritative guard returns the documented 20002.

cashu/mint/ledger.py
```diff
@@ -685,7 +691,7 @@ def create_internal_melt_quote(
         if not mint_quote.method == method.name:
             raise TransactionError("methods do not match")
         if mint_quote.paid:
-            raise TransactionError("mint quote already paid")
+            raise InvoiceAlreadyPaidError("mint quote already paid")
```
callebtc, line 694: Please carry the mapping through to the immediately following `mint_quote.issued` branch: it still raises generic TransactionError (11000), while the documented code for an issued mint quote is QuoteAlreadyIssuedError (20002). The same unconverted branch also remains in melt_mint_settle_internally.

cashu/mint/db/write.py
KvngMikey: great catch, has been handled.

cashu/mint/ledger.py
KvngMikey: fixed

a1denvalu3:
---
title: "NPC wallet marks unpaid/unissued quotes as ISSUED because it dispatches on the wrong mint error code (11000 instead of 20002)"
slug: npc-mint-error-code-misdispatch
date: 2026-09-02
status: confirmed
severity: medium
finding: true
relation: revealed
target: [cashubtc/nutshell]
nuts: [NUT-04]
---

## Summary

PR cashubtc/nutshell#1129 ("fix/nut-error-code-alignment") is a mint-side
refactor that aligns error classes/codes with the NUT error-code registry.
Reviewing exactly which code the mint now returns for each mint failure reveals
a latent wallet-side bug in `cashu/wallet/npc.py` (untouched by the PR): the
npub.cash NPC minting loop detects "quote already issued" by matching the
hardcoded string `"Code: 11000"`, but 11000 is the *generic TransactionError*
code, while the mint returns **20002** (`QuoteAlreadyIssuedError`) for the real
already-issued case. The check therefore (a) never matches the condition it was
written for (false negative), and (b) matches unrelated mint failures that
carry code 11000 (false positive), after which the NPC permanently marks the
local quote as ISSUED and never retries minting it — stranding a paid Lightning
invoice with no ecash issued.

## Root Cause

`cashu/wallet/npc.py:189-199`:

```python
except Exception as e:
    # If the mint returns an error that the quote is already issued, we assume it is issued
    if "Code: 11000" in str(e) or (isinstance(e, CashuError) and e.code == 11000):
        print(f"Quote {quote_id} already issued (mint). Updating local state.")
        await update_bolt11_mint_quote(
            db=self.wallet.db,
            quote=quote_id,
            state=MintQuoteState.issued,
            paid_time=int(time.time()),
        )
        continue
```

The mint surfaces errors as `{"detail": ..., "code": ...}`
(`cashu/mint/app.py` exception middleware, ~lines 97-105), and the wallet raises
`Exception(f"Mint Error: {detail} (Code: {code})")`
(`cashu/wallet/v1_api.py:137-140`), so `str(e)` contains the mint's numeric
code.

The PR's error-code tables (`cashu/core/errors.py`) show why the dispatch is
wrong:

- `TransactionError` (the generic fallback) has code **11000**
  (`cashu/core/errors.py:30-35`). Mint failures that are *not* "already issued"
  but return 11000 include "amount to mint does not match quote amount" and
  "quote unit does not match output unit" (`cashu/mint/ledger.py:528-531`).
  Pre-PR this list additionally included "Mint quote already pending." and
  "quote expired" — the PR re-mapped those to 20005/20007, which narrows but
  does not eliminate the false-positive surface.
- `QuoteAlreadyIssuedError` has code **20002** (`cashu/core/errors.py:220-225`)
  and is what `ledger.mint()` actually raises for an issued quote
  (`cashu/mint/ledger.py:520-521`). The npc check never matches it.

So the "already issued" self-healing path in npc.py is dead code for its
intended purpose and a trap for unrelated 11000 failures.

## Insight gained from this PR (relation: revealed)

`cashu/wallet/npc.py` is not modified by this PR. The bug was discovered
directly from the PR's changes: the PR enumerates and remaps the exact NUT
error codes the mint returns on every mint-failure branch (11000/20001/20002/
20005/20007/20008/11003/11004). Auditing "which code does each `ledger.mint()`
failure now produce" against every consumer of those codes surfaced that the
only code-based consumer in the codebase, `npc.py:191`, dispatches on the
generic 11000 while the condition it wants ("already issued") is 20002 —
before and after the PR. The PR's alignment work makes this mis-dispatch
explicit (it even re-maps two of the false-positive triggers), but the
underlying wallet bug is pre-existing and remains exploitable post-PR.

## Attack Steps / Reproduction

1. Set up the npub.cash NPC flow (`NpubCash.mint_quotes()` polls paid quotes
   and mints them via `wallet.mint()`).
2. Cause any mint attempt for a paid, un-issued quote to fail with a
   11000-coded error (post-PR: e.g. amount/unit mismatch; pre-PR also "quote
   expired" / "Mint quote already pending.").
3. `npc.py` matches `"Code: 11000"`, marks the quote `issued` in the local
   wallet DB, and `continue`s.
4. On every subsequent poll the local quote is `issued`, so NPC prints
   "already minted" and skips it forever. The invoice was paid, but no ecash
   was ever issued to the user and the automated flow has no recovery path.

PoC executes the real `NpubCash.mint_quotes()` code path with a stubbed wallet:

- Case 1 (false positive): `wallet.mint` raises
  `Exception("Mint Error: amount to mint does not match quote amount (Code: 11000)")`
  → npc calls `update_bolt11_mint_quote(..., state=MintQuoteState.issued)`
  even though nothing was issued.
- Case 2 (false negative): `wallet.mint` raises
  `Exception("Mint Error: quote already issued (Code: 20002)")`
  (the genuine `QuoteAlreadyIssuedError` response) → npc does NOT update the
  local state and will keep retrying.

## Impact

Wallet-side loss-of-funds availability in the automated NPC/npub.cash flow: a
paid quote is irreversibly (from the NPC's perspective) marked as issued when
minting failed, so the user pays a Lightning invoice but never receives the
corresponding ecash, with no automatic retry or recovery. Additionally, the
intended idempotency handling for genuinely-issued quotes never fires, causing
endless failed retries. The mint side is unaffected.

## Test Results

Executed against the PR branch (Poetry env):

```
$ poetry run python poc_npc_error_code.py
Quote quote-1 already issued (mint). Updating local state.
[false-positive / unrelated 11000 error] exception: Mint Error: amount to mint does not match quote amount (Code: 11000)
    quote marked ISSUED locally: True
Failed to mint quote quote-1: Mint Error: quote already issued (Code: 20002)
[false-negative / real already-issued 20002] exception: Mint Error: quote already issued (Code: 20002)
    quote marked ISSUED locally: False

PoC SUCCESS:
 - an unrelated 11000-coded mint failure falsely marks the
   local quote as ISSUED (strands the paid invoice)
 - the genuine 'already issued' response (20002) is NOT detected
```

## Proposed Fix

In `cashu/wallet/npc.py`, dispatch on the correct NUT code for
"quote already issued" instead of the generic transaction error code, and
prefer structured error handling over string matching:

```python
except Exception as e:
    code = e.code if isinstance(e, CashuError) else None
    if code is None:
        # parse "Mint Error: <detail> (Code: <n>)" raised by v1_api
        m = re.search(r"\(Code: (\d+)\)", str(e))
        code = int(m.group(1)) if m else None
    if code == 20002:  # QuoteAlreadyIssuedError
        await update_bolt11_mint_quote(
            db=self.wallet.db,
            quote=quote_id,
            state=MintQuoteState.issued,
            paid_time=int(time.time()),
        )
        continue
    print(f"Failed to mint quote {quote_id}: {e}")
```

Even better: make `cashu/wallet/v1_api.py::raise_on_error_request` raise a
`CashuError` subclass carrying the mint's numeric code (it already parses
`resp_dict["code"]`) so consumers never need to regex the message string, and
audit other places (e.g. CLI TODO at `cashu/wallet/cli/cli.py:626`) to consume
that structured code.

PR #1131, opened 2026-08-23
feat: add payment-method abstraction

a1denvalu3:
## Summary

- add a typed, allowlisted payment-method plugin registry with BOLT11 as the built-in adapter
- generalize mint quote, melt quote, settlement, feature discovery, wallet APIs, persistence, and CLI method selection
- retain the existing `/v1/.../bolt11` endpoints, defaults, wire fields, and legacy wallet tables for older clients
- add method-specific metadata storage and safe response projection so internal quote fields are never exposed as plugin extras

This establishes the abstraction boundary for methods such as BOLT12 or on-chain payments without adding or advertising a new production payment rail.

## Compatibility

- existing BOLT11 routes and public wallet methods remain available
- BOLT11 remains the default method
- legacy BOLT11 quote tables remain canonical for existing clients
- new database changes are additive
- method-specific request/response data is validated by the selected plugin

## Validation

- BOLT11 mint/melt: 41 passed, 1 skipped
- wallet: 32 passed, 1 skipped
- payment-method regression tests: 7 passed
- Ruff: passed
- `git diff --check`: passed

`make check` still reports two unrelated existing Optional-path mypy errors in `cashu/mint/management_rpc/management_rpc.py`.

## Review notes

This includes additive schema and protocol-boundary work, so it requires maintainer review under the repository contribution policy.


cashu/mint/ledger.py
```diff
@@ -734,12 +785,27 @@ async def get_mint_quote(self, quote_id: str) -> MintQuote:
                     )
                     if not quote:
                         raise Exception("quote not found")
-                    if quote.unpaid:
-                        logger.trace(f"Setting quote {quote_id} as paid")
-                        quote.state = MintQuoteState.paid
-                        quote.paid_time = now
+                    total_paid = reported_paid + quote.amount_paid_internal
```
orangeshyguy21, line 788: Internal settlement credits `amount_paid_internal` without paying or cancelling the bolt11 invoice, so the invoice stays payable. 

If anyone pays it afterwards, this adds `quote.amount` on top of the internal credit, flips the quote back to PAID, and `mint()` then allows a second issuance.

For non-reusable methods should double payment continue to be refuted?

cashu/mint/migrations.py
```diff
@@ -1277,6 +1277,20 @@ async def m036_add_mint_quote_accounting_fields(db: Database):
         await conn.execute(
             f"ALTER TABLE {db.table_with_schema('mint_quotes')} ADD COLUMN updated_at TIMESTAMP DEFAULT NULL"
         )
+        await conn.execute(
```
orangeshyguy21, line 1280: This backfill gets skipped for mints on 20.3 and above. 
Is this intentional, or should it be a new migration?

cashu/mint/crud.py
```diff
@@ -616,8 +618,8 @@ async def store_mint_quote(
         await (conn or db).execute(
             f"""
             INSERT INTO {db.table_with_schema("mint_quotes")}
-            (quote, method, request, checking_id, unit, amount, state, created_time, paid_time, issued_time, last_checked, pubkey, amount_paid, amount_issued, updated_at)
-            VALUES (:quote, :method, :request, :checking_id, :unit, :amount, :state, :created_time, :paid_time, :issued_time, :last_checked, :pubkey, :amount_paid, :amount_issued, :updated_at)
```
orangeshyguy21, line 620: This insert has no duplicate-request check and the table has no uniqueness on `request` or `checking_id`, so two live quotes can carry the same payment request. 

For methods with `allows_repeated_payments`, each quote independently sums whatever the processor reports for that request, so one payment can fully fund both quotes. It also makes the request-based lookups in internal settlement (`ledger.py:1331`, `melt_mint_settle_internally`) ambiguous, crediting an arbitrary matching quote.

cashu/wallet/v1_api.py
```diff
@@ -371,21 +372,43 @@ async def mint_quote(
...
+        method_options: Optional[dict] = None,
+    ) -> PostMintQuoteResponse:
+        """Request a mint quote for an explicitly selected payment method."""
+        logger.trace(f"Requesting mint: POST /v1/mint/quote/{method}")
         payload = PostMintQuoteRequest(
             unit=unit.name, amount=amount, description=memo, pubkey=pubkey
         )
+        payload_dict = payload.model_dump()
+        if method_options:
+            payload_dict.update(method_options)
```
orangeshyguy21, line 397: `payload_dict.update(method_options)` runs after the standard payload is built, so a `--method-options` JSON key like `quote` or `amount` silently overwrites the normative field previously set. 

Should colliding keys be rejected here?

orangeshyguy21 (changes requested):
Reviewed the changes here. 

I left some inline comments on accounting and balance preservation that deserve some thought, as well as a few open questions.

PR #999, opened 2026-05-05
feat(crypto): migrate BDHKE to BLS12-381 (v3 keysets)

a1denvalu3:
>[!CAUTION]
>This PR bumps Nutshell's version to 0.21

## Summary
This pull request introduces BLS12-381 cryptography into the Cashu protocol, enabling smaller proofs and paving the way for multi-signature schemes and batch verification.

### Core Changes
- Introduces `v3` keysets using the BLS12-381 curve.
- Adds multiplicative blinding logic (`Y * r`) to replace legacy additive blinding (`Y + r*G`).
- Replaces DLEQ proof requirements with BLS pairing verification (`e(C, G2) == e(Y, K2)`).
- Modifies wallet redemption and proof construction steps to unblind signatures cleanly and omit unneeded DLEQ verification logic.
- Maintains complete backwards compatibility with `v1`/`v2` (`secp256k1`) keysets.
- Implements comprehensive batch pairing verification for unblinded BLS signatures.
- Updates keyset ID generation format to utilize the `02` prefix for BLS keysets.

### Testing
- Updates the wallet and mint CLI tests to dynamically accommodate both `secp256k1` and `BLS12-381` logic.
- Adds `tests/test_crypto_bls.py` test suite specifically for deterministic hash-to-curve testing, verification of individual BLS protocol steps, and batched BLS pairing checks.

PR #1134, opened 2026-08-24
feat: version 02 BATs sign the request they authorize (NUT-22)

robwoodgate:
## Intended to merge into BLS Keyset v3 PR  #999 

Implements nutroot blind authentication per cashubtc/nuts#421's NUT-22 section: on a v3 (version 02) auth keyset a BAT's secret is a point and its AuthProof carries a required key-path witness over the request transcript.

**Core:** the transcript module gains the `0x05` authorized-request container (`01` method, `02` origin-form target, `03` body hash) under the same domain tag and digest rule as every transcript; `nut22.py` holds the sign/verify helpers. Pinned byte for byte to the nuts `tests/22-tests.md` vector, including the vector's witness signature.

**Mint:** the blind-auth middleware passes method, target and body into `verify_blind_auth` (the body is read only when the token peeks as version 02; Starlette caches the read for the downstream app). On a BLS auth keyset the mint enforces: point secret, witness present, exactly one signature, valid over this request's digest. Pre-v3 BATs are unchanged.

**Wallet:** v3 BATs mint as fresh random keypairs (secret = pubkey), each private key kept with its proof as a `BATKEY:` derivation record in the auth db; at presentation the wallet pre-serializes the body so the signed bytes are the sent bytes, derives the target from the full URL (subpath-mounted mints verify their real path), and attaches the witness to the serialized BAT.

**Tests:** vector-pinning unit tests (`tests/test_blind_auth_v3.py`), a `verify_blind_auth` accept/reject case for v3 BATs, and updated middleware/server unit mocks for the new signature. Gates locally: ruff, mypy, full pytest 1022 passed.

The cashu-ts side ships the matching client in cashubtc/cashu-ts#950.
Cross-refs: cashubtc/nuts#421 (spec)


robwoodgate:
Pushed ee5e839 after running the full loop end to end via cashu-ts's auth example (Keycloak + this mint from source): CAT login, v3 BAT minting on the BLS auth keyset, then BAT-authorized quote/mint/swap/receive with every protected request's witness verified.

Note for the realm side: Keycloak 24+ only issues the sub claim with the basic client scope granted, which the cashu-ts example realm now does.

a1denvalu3 (approved):

a1denvalu3 (approved):

PR #1137, opened 2026-08-25
test: sync the shared nutroot vectors with the spec

robwoodgate:
## Intended to merge into BLS Keyset v3 PR  #999

Test data only, no code changes.

The shared `nutroot_v3_vectors.json` `melt_with_change` transcript entry predates the NUT-13 reframing (cashubtc/nuts `f5353a0`): `swap`, `mint` and `melt` were regenerated for the new counter-0 secret but this entry kept the old one, so nutshell and cashu-ts agreed with each other while both disagreed with the spec document (`tests/10-tests.md` pins digest `3b7a268b...`; the JSON carried `98bd69...`).

This regenerates the entry from the spec and adds two vectors the spec carries but the shared JSON did not: the duplicate-pair fold (the fold commits the leaf multiset) and the unknown-leaf-field rejection. `test_nutroot.py` recomputes the transcript and digest from each vector's `tx`, so the suite now proves the encoder reproduces the spec values byte for byte. The cashu-ts copy is corrected in step on its `rnd/taproot-v3` branch.


a1denvalu3:
Already aligned in the main PR

PR #1139, opened 2026-08-31
Nutroot: per-input signing digests, disclosure, and NUT-07 spend commitments

robwoodgate:
# Intended to merge into #999

Syncs the mint with the revised NUT-10/NUT-07 design on cashubtc/nuts#443, stacked on #999.

**Signing.** A transaction keeps one shared transcript, and each v3 input signs its own input digest: `tagged_hash("Cashu_TransactionInput", transaction_digest || SHA256(its own container record))`. The mint verifies per input; quote locks (single and NUT-29 batch) sign their own quote input's digest. A witness can no longer verify at any input but its own, and the transcript builder rejects a repeated proof or quote input. The melt quote output binds the quote's `amount` plus its `fee_reserve`, what the quote is entitled to draw from the inputs, on the mint and wallet sides alike.

**Caps.** Leaf bodies at 512 bytes, trees at 8 leaves (paths of at most 3 sibling hashes), serialized witnesses at 4096 characters, per the updated NUT-10.

**Disclosure and NUT-07.** Leaves gain the `disclosure` field (`0x0a`, mode `0x01` only; other values fail closed). A spent v3 proof's checkstate entry carries a spend commitment, `tagged_hash("Cashu_SpendCommitment", Y || input_digest || SHA256(witness))`, derived at serve time from what the mint already stores; the exact witness string and its input digest are served only when the exercised leaf carries disclosure. NUT-17 initial snapshots and live events go through the same serving helper. The stored `digest` column now holds the input digest, so there is no schema change.

**Commit leaf.** NUT-10 allocates leaf type `0x04`, `commit`: one 32-byte hash and no spend rule, binding a tree to outside context (NIP-61 replay protection is the first use). The registry names it, the serializer and parser accept exactly that shape, and `verify_script_path_spend` refuses a witness that reveals one. The mint only ever sees it as a sibling hash, so nothing changes for spends through other leaves. Unallocated types now start at `0x05`, so the illustrative `melt_to` example moves there.

**Inputs by Y.** Field 03 of a proof input container is `Y`, the keyset's hash_to_curve of the secret, on every keyset version (NUT-10): 48 bytes under a v3 keyset, 33 under a pre-v3 one in a mixed transaction. A secret never enters a transcript, which is what a co-signer or an auditor is shown. `TranscriptProofInput` takes `Y`, contexts key by it, and the mint and wallet hash each input once via `proof_transcript_y`.

**Vectors.** `tests/nutroot_v3_vectors.json` matches the canonical cashu-ts copy, including a two-input transcript pinning distinct input digests, a mixed-keyset transcript with a pre-v3 input named by its secp256k1 Y, the auditable-lock disclosure example with its commitment opening, and the commit-leaf tree.

**Verification.** ruff, mypy, full pytest locally (1018 passed, 52 skipped; regtest is CI's), and the cashu-ts integration suite (80/80, 32 v3 + 48 legacy) against a mint built from this branch.


robwoodgate:
Had to re-trigger CI a couple of times, seems the retests can be a bit flakey - have raised an issue: https://github.com/cashubtc/nutshell/issues/1140

PR #1145, opened 2026-09-04
fix(auth): align auth promises schema with the mint CRUD

### LINKED ISSUE #1143: Bug: NUT-22 blind-auth broken: auth DB schema missing
### Blind-auth minting fails on a fresh, officially migrated auth DB

`POST /v1/auth/blind/mint` fails with:

```
sqlite3.OperationalError: table promises has no column named mint_quote
```

The auth ledger stores blinded messages via the shared `LedgerCrudSqlite.store_blinded_message` (`cashu/mint/crud.py`), which inserts into `promises(… mint_quote, melt_quote, swap_id, order_index)`. The auth migrations (`cashu/mint/auth/migrations.py`, `m000`–`m004`) create `promises` without those four columns. The main-ledger migrations add them, the auth ones were never updated to match.

Version: 0.20.3 (`docker.io/cashubtc/nutshell`). Regtest lab: LND 0.20.0-beta, Bitcoin Core 30.0, Keycloak 25.0.6 as the NUT-21 OIDC provider.

KvngMikey:
## Summary
The auth ledger uses the shared LedgerCrudSqlite, which writes mint_quote, melt_quote, swap_id, order_index and signed_at, and stores blinded messages before signing them. The auth migration chain never grew those columns and still declares c_ NOT NULL, so blind auth minting fails on a freshly migrated auth database with:
```
    table promises has no column named mint_quote
```
Add m005 to add the missing columns and make c_ nullable.

### DIFF
--- a/cashu/mint/auth/migrations.py
+++ b/cashu/mint/auth/migrations.py
@@ -138,3 +138,59 @@ async def m004_remove_dleq_from_promises(db: Database):
         await conn.execute(
             f"ALTER TABLE {db.table_with_schema('promises')} DROP COLUMN dleq_s"
         )
+
+
+async def m005_align_promises_with_mint_schema(db: Database):
+    """
+    Align the auth promises table with the columns the shared LedgerCrudSqlite writes.
+    """
+    async with db.connect() as conn:
+        if conn.type == "SQLITE":
+            await conn.execute("PRAGMA foreign_keys=OFF;")
+            await conn.execute(
+                f"""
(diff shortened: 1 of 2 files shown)


a1denvalu3:
@KvngMikey `PRAGMA foreign_keys=ON;` is ignored by sqlite because it's inside an active transaction. This migration therefore disables foreign keys

KvngMikey:
> @KvngMikey `PRAGMA foreign_keys=ON;` is ignored by sqlite because it's inside an active transaction. This migration therefore disables foreign keys

@a1denvalu3  you're correct, i've updated this now.

PR #1148, opened 2026-09-10
ci: pin regtest setup and update fee-rounding test

a1denvalu3:
The regtest workflows follow upstream's default branch, so the test network can change independently of Nutshell. Pin the `callebtc/cashu-regtest` checkout to `7eb3515b9fb41f477f831cedd28e1ad9ae6c521d`, a tested revision with CLN 26.06.7 and support for `xpay`.

CI still checks out the regtest repository and runs `start.sh` to initialize the network and channels and run upstream acceptance checks before the Nutshell tests. Image pulling is handled by the upstream script, which uses published service images where available and can fall back to source builds.

Changes:

- Replace the unpinned `git clone` in both workflows with an explicit checkout of the pinned revision.
- Always tear down the `cashu` Compose project after each regtest job, including its network and volumes.
- Increase the mint timeout from 10 to 30 minutes and the wallet timeout from 10 to 20 minutes to allow setup and tests to finish on slower runners.
- Update the fee-rounding test and its invoice helper for the expanded network. A cheaper route can charge an exact whole-satoshi fee, invalidating the old fixed-fee assumption. Use the isolated fee-testing node, require a fractional-satoshi fee, and verify rounding and returned change against LND's settled fee.

Validation:

- [Full CI run](https://github.com/cashubtc/nutshell/actions/runs/34493665695) passed on `672120a1`: all 21 jobs, including all 12 regtest matrix jobs.
- Local mint API tests passed with LND REST and LND RPC on SQLite and PostgreSQL: 88 passed, 4 skipped.
- `make format` and `make check` passed. Upstream startup checks and removal of the Compose containers, network, and volumes were also verified locally.

### DIFF
--- a/.github/workflows/regtest-mint.yml
+++ b/.github/workflows/regtest-mint.yml
@@ -29,7 +29,8 @@ env:
 jobs:
   regtest-mint:
     runs-on: ${{ inputs.os-version }}
-    timeout-minutes: 20
+    # Allow startup plus the full mint suite on slower runners.
+    timeout-minutes: 30
     steps:
       - name: Start PostgreSQL service
         if: contains(inputs.mint-database, 'postgres')
--- a/.github/workflows/regtest-mint.yml
+++ b/.github/workflows/regtest-mint.yml
@@ -23,10 +23,8 @@ permissions:
   contents: read
 
 env:
-  # Prebuilt cashu-regtest stack (bitcoind, 3x LND, 3x CLN, LNbits). Pinned to
-  # a tag whose docker-compose exposes the node ports on the host (3010, 3001,
-  # 8081, 10009, 5001) with RPC ports consistent with its own scripts.
-  REGTEST_IMAGE: ghcr.io/callebtc/cashu-regtest-environment:sha-bde3931fe27e32b8f83b2adbb65a8460c2f32ef5
+  # This revision uses CLN 26.06.7 (with xpay) and pulls prebuilt service images.
+  REGTEST_REF: 7eb3515b9fb41f477f831cedd28e1ad9ae6c521d
 
 jobs:
   regtest-mint:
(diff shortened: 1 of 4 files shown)


a1denvalu3:
Status update: the image-based regtest setup works — the stack bootstraps in about a minute and all LND-backed jobs are green. The remaining failures are the 4 `CLNRestWallet` jobs, and they're blocked on the published image being stale, not on this PR's wiring:

- Every published tag of [`ghcr.io/callebtc/cashu-regtest-environment`](https://github.com/callebtc/cashu-regtest/pkgs/container/cashu-regtest-environment) was built 2025-05-11 and ships `boltz/c-lightning:24.08`.
- Nutshell's `CLNRestWallet` now pays via CLN's `xpay` plugin (#1127, `POST /v1/xpay`), which requires Core Lightning ≥ 25.02. On the published image every CLN payment fails with `Unknown command 'xpay'`, so melts stay `PENDING` and the CLN assertions fail.

@callebtc could you publish a fresh `cashu-regtest-environment` image from current [`callebtc/cashu-regtest`](https://github.com/callebtc/cashu-regtest) `main` (which pins CLN v26.06.7)? Once a new tag is up, this PR only needs a one-line bump of `REGTEST_IMAGE`.

Two bugs in the existing tags worth avoiding in the next publish (both break host-side consumers):

1. `sha-e4a2558...` remaps lnd-2 to `--rpclisten=lnd-2:10010` while `docker-scripts.sh` still dials `lnd-2:10009` → bootstrap hangs forever in `waiting for lnd-2 to sync`.
2. `sha-6160cb1...` (the package's current latest) has no `ports:` section in its compose file → no node ports reach the host at all.

This PR is currently pinned to `sha-bde3931...`, the only existing tag that is both port-published and self-consistent.

Issue #1152, opened 2026-09-13
LND backends: an invoice amount that is not a whole number of sats is paid as a partial payment and the melt gets stuck pending


gorrdy:
## Summary

On both LND backends, `pay_invoice()` decides whether to make a *partial* (MPP) payment by comparing the invoice amount with the melt quote amount. For a `sat`-unit backend the quote amount is rounded **up** to whole sats, so an invoice denominated with sub-satoshi precision produces a mismatch and is misclassified as a partial payment — even though no wallet requested MPP.

The partial path then resolves the route with `QueryRoutes`, which is called **without the invoice's route hints**. A destination reachable only through a private channel — Phoenix, Wallet of Satoshi, and most mobile wallets — does not exist in the public graph, so no route is found, no HTLC is sent, and no payment is registered with lnd.

Since the backend has no record of the payment, the mint cannot conclude that it failed. The melt quote stays `PENDING` and the payer's proofs stay locked.

## Impact

The payer's ecash is locked indefinitely. They cannot retry (a new quote for the same invoice is refused while one is pending), cannot spend the proofs (`ProofsArePendingError`), and the quote's expiry releases nothing. Only the mint operator can unlock it.

Fiat-denominated invoices routinely carry sub-satoshi amounts, so the trigger is ordinary: **a user pays a mobile wallet an amount entered in their local currency**. On one production mint this was 1 of 684 melts over 30 days — uncommon, but it failed every time it occurred.

This may also be what [#759](https://github.com/cashubtc/nutshell/issues/759) ran into: payments from the mint to Phoenix and Wallet of Satoshi failed with `no_route`, while the same payments made directly from the same node succeeded, and payments from the mint to a public node worked. That issue was closed after the reporter raised `LIGHTNING_FEE_PERCENT` as a workaround; the cause was never identified.

## Root cause 1 — an amount mismatch is used as the MPP signal

`cashu/lightning/lnd_grpc/lnd_grpc.py:167`, identical at `cashu/lightning/lndrest.py:199`:

```python
if amount_msat != quote_amount.to(Unit.msat).amount and self.supports_mpp:
    return await self.pay_partial_invoice(quote, quote_amount, fee_limit_msat)
```

`get_payment_quote()` rounds the quote amount up to the backend unit (`lnd_grpc.py:456`, `round="up"`), so an invoice of `24_779_417` msat yields a quote of `24_780` sat = `24_780_000` msat. The values differ by 583 msat and the branch is taken.

[...]

PR #1153, opened 2026-09-13
fix(lightning): don't treat sub-satoshi rounding as a partial payment

gorrdy:
Fixes root cause 1 of #1152.

## Problem

`pay_invoice()` decides whether to make a partial (MPP) payment by comparing the invoice amount with the melt quote amount. `get_payment_quote()` rounds the quote amount **up** to the backend unit, so an invoice denominated with sub-satoshi precision leaves the quote slightly *above* the invoice amount, and the inequality treats that rounding as a request for a partial payment.

An ordinary melt is then sent through `pay_partial_invoice`, which resolves the route with `QueryRoutes`. `QueryRoutes` only knows the public graph, so a destination published in the invoice's route hints — Phoenix, Wallet of Satoshi, most mobile wallets — is not found. No HTLC is sent and no payment is registered with lnd, which leaves the melt quote pending and the payer's proofs locked.

Sub-satoshi amounts are what a wallet produces when the user enters an amount in fiat, so this is reachable from an ordinary payment and does not need an MPP-capable wallet to trigger it. On one production mint it was 1 of 684 melts over 30 days, and it failed every time it occurred.

## Change

Take the partial path only when the invoice asks for more than this quote covers, which is the case a partial payment actually describes:

```python
-if amount_msat != quote_amount.to(Unit.msat).amount and self.supports_mpp:
+if amount_msat > quote_amount.to(Unit.msat).amount and self.supports_mpp:
```

Applied to both `LndRPCWallet` and `LndRestWallet`, which carried the same condition.

A genuine MPP quote pays a part of the invoice, so the invoice amount is larger; sub-satoshi rounding puts the quote amount larger, by less than 1000 msat. The two cases sit on opposite sides of the comparison, so the direction is enough to separate them without persisting new state.

## What this does not fix

`QueryRoutes` is still called without the invoice's route hints (root cause 2 of #1152), so genuine MPP melts to a private-channel destination remain broken. That is a separate change and I have no way to test it against a real private-channel destination.

One degenerate case also stays misclassified: a partial payment within one satoshi of the invoice total would round up to the full amount and be paid in full. Persisting `mpp_amount` on the melt quote and branching on the actual request would remove the guesswork entirely, at the cost of a migration.

## Tests

`tests/lightning/test_lnd_subsat_invoice.py` covers both backends: a sub-satoshi invoice must be paid in full with the whole payment request handed to lnd, and a quote below the invoice amount must still use the partial path. The first two fail without the change; the last two guard the MPP behaviour that must not change.

Locally (Python 3.10):

- `pytest tests` — 970 passed, 52 skipped, 1 failed. The failure is `test_melt_preserves_change_signatures_order_integration`, which shells out to `docker exec cashu-lnd-1-1 lncli`; it fails identically on unmodified `main` @ `6e77e85` in the same environment, since regtest is not running.
- `make ruff-check` — clean
- `mypy cashu --check-untyped-defs` — no new errors (the 4 in `cashu/core/base.py` are present on `main`; that file is untouched here)

I have not run the regtest suites.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


callebtc:
thank you, great fix!

PR #1169, opened 2026-09-19
fix(mint): refuse the .env.example placeholder as a seed

btcforplebs:
`.env.example` ships:

```
# MINT_PRIVATE_KEY=<openssl rand -hex 32>
```

If that line is uncommented and the placeholder is pasted literally instead of being replaced by the command's output, the mint starts normally and derives its keyset from the string `<openssl rand -hex 32>`. That seed is public, so anyone can sign valid ecash for the mint. The existing startup checks only catch an empty key and the exact string `supersecretprivatekey`.

This adds the placeholder to the same rejection, so the mint refuses to start with either known string. Three lines changed, no new lines.

Verified: `MintKeyset(seed="<openssl rand -hex 32>", ...)` and `MintKeyset(seed="supersecretprivatekey", ...)` both raise; a real 64-hex seed still constructs. `tests/mint/test_mint_init.py` and `tests/wallet/test_wallet_keysets_v2.py` pass (33 passed, 5 skipped). Ruff 0.7.4 and mypy clean on the file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ye0man (approved):
Ack

KvngMikey (commented):
lgtm

KvngMikey (approved):

btcforplebs:
Thank you

PR #1171, opened 2026-09-20
fix(crypto): mask public-key scalar multiplication

a1denvalu3:
Nutshell passes secret scalars into Coincurve's variable-time public-key multiplication. This PR masks the secret in `P * key`: choose a fresh random scalar $t$ and compute $((s+t) \bmod n)P - tP = sP$, where $n$ is the group order. P2BK ECDH uses this operator too.

The implementation uses Coincurve directly: `PrivateKey()` generates $t$, `.add()` computes the masked scalar, and `.multiply()` computes both points. Invalid masks and zero masked sums are retried; RNG failures propagate. Inputs and the point's context are preserved, and Coincurve's `.multiply()` stays unchanged.

The cost is one extra arbitrary-point multiplication, subtraction, and mask generation; constructing private-key objects also computes their public keys. This is a timing mitigation, not a constant-time guarantee.

Timing rationale:

Consider repeated calls with the same secret $s$ and point $P$. In this simplified model, $t$ is uniform modulo $n$ and $u=(s+t)\bmod n$. We ignore rejected masks and setup/subtraction costs. Let $f(x)$ be multiplication time and $\varepsilon$ independent, zero-mean noise. The observer sees only the total duration:

```math
R_s = f(t) + f(u) + \varepsilon.
```

Adding $s$ modulo $n$ just rearranges the possible mask values: both $t$ and $u$ remain uniform, whatever $s$ is. Their timings therefore have the same mean $\mu$:

```math
\mu = \mathbb{E}[f(t)] = \mathbb{E}[f(u)].
```

Without masking, the mean duration in this model is $f(s)$. With masking, expectations add, even when the two times are correlated:

```math
\mathbb{E}[R_s]
= \mathbb{E}[f(t)] + \mathbb{E}[f(u)] + \mathbb{E}[\varepsilon]
= \mu + \mu + 0 = 2\mu.
```

The mean no longer depends on $s$ in this model.

Now consider variance: the spread of timings around their mean. Both multiplication times have the same variance $\sigma^2$; let $\sigma_{\varepsilon}^2$ be the noise variance. Since $u-t\equiv s\pmod n$, the two multiplication times may tend to rise or fall together. Their covariance $C(s)$ measures that relationship:

```math
\begin{aligned}
\sigma^2 &= \mathrm{Var}[f(t)] = \mathrm{Var}[f(u)], \\
\sigma_{\varepsilon}^2 &= \mathrm{Var}[\varepsilon], \\
C(s) &= \mathrm{Cov}[f(t), f(u)].
\end{aligned}
```

The variance of a sum includes twice the covariance; independent noise adds only its own variance:

```math
\begin{aligned}
\mathrm{Var}[R_s] &= \mathrm{Var}[f(t)] + \mathrm{Var}[f(u)] + 2C(s) + \mathrm{Var}[\varepsilon] \\
&= 2\sigma^2 + 2C(s) + \sigma_{\varepsilon}^2.
\end{aligned}
```

For $N\geq2$ measured durations $R_{s,i}$, the sample mean $\bar{R}$ and sample variance $\hat{v}$ estimate the total timing mean and variance:

```math
\bar{R} = \frac{1}{N}\sum_{i=1}^{N}R_{s,i}, \qquad
\hat{v} = \frac{1}{N-1}\sum_{i=1}^{N}(R_{s,i}-\bar{R})^2.
```

Rearranging the variance equation gives an estimate $\widehat{C}(s)$ of the covariance, **if** the multiplication and noise variances have been separately estimated:

```math
\widehat{C}(s) := \frac{\hat{v}-2\sigma^2-\sigma_{\varepsilon}^2}{2}.
```

The practical difficulty is separating these contributions. More timings improve $\hat{v}$, but do not by themselves tell us how much comes from multiplication, noise, or covariance. Calibration must match the target's timing conditions: errors in either baseline variance can look like covariance. If differences in $C(s)$ are small, sampling and calibration errors must be smaller still, potentially requiring many measurements and stable conditions.

Even an accurate covariance estimate is not a recovered key: an attacker still needs an efficient way to connect it to secret bits. These are obstacles to an attack, not proof of impracticality; we have not demonstrated exploitable covariance or practical key recovery.

Regression tests cover equivalence with Coincurve across varied points and boundary scalars, fresh masks, invalid masks, zero-sum retries, modular wraparound, input-type validation, RNG failure, input/context preservation, P2BK routing, and Coincurve's unchanged direct multiplication behavior.

Validation:

- `make format`
- `make check` (Ruff and mypy)
- `pytest -o addopts= -q tests/test_secp.py tests/test_crypto.py tests/wallet/test_wallet_p2bk.py tests/mint/test_mint.py` — 81 passed with FakeWallet on Python 3.12 and isolated test databases.


a1denvalu3:
Empirical follow-up on the timing model in this PR's description. I mounted the mean-based attack against the unmasked path and measured the covariance channel against the masked path. Diagnostics: `security-reviews/secp256k1-timing/kocher_unmasked_attack.py`, `masking_variance_attack.py`, `masking_variance_reanalysis.py` (raw samples in `masking_variance_data.npz`).

**Scope caveat:** all measurements are local, single-machine, single-session, on the native path via coincurve — unmasked timings call `PublicKey.multiply()` directly; masked timings call the patched `P * priv` wrapper. No HTTP/application stack, so the unmasked numbers are attacker-favorable upper bounds.

## Attack on unmasked multiplication (works)

The model comes from `ecmult_impl.h` (libsecp256k1 0.5.0): the scalar is GLV-split into two ~128-bit halves, each wNAF-recoded with window 5; the main loop runs one doubling per position and one addition per nonzero digit. So `T(s) ≈ β0 + βz·z(s) + βh·h(s)`, where `z(s)` = total nonzero wNAF digits and `h(s)` = loop length. I reimplemented `split_lambda` + the wNAF recoding in Python (validated against the group law) to build the attacker's prediction model.

- Fit on 3000 random scalars: `βz ≈ 108 ns/digit`. `z(s)` over random scalars: mean 43.2, std 1.6, range ~38–48. The z-term accounts for ~175 ns of secret-dependent spread, matching an independent ANOVA estimate of the total signal (~200 ns std).
- Attack: 100 unknown secrets, fresh attacker-chosen point per sample, N total-time samples per secret, invert the model on the sample mean. Result: **corr(z_true, z_est) ≈ +0.6–0.67 at N = 100–5000**, MAE ≈ 1.7 digits, |z_est − z| ≤ 1 in 52–64% of cases.

So the first-order (mean) channel is real and exploitable at ~10²–10³ samples. Honest limits: what leaks is ~0.4 measured bits (channel capacity ~2.5–3 bits) of *aggregate* information per fixed secret — the wNAF recoding's carries destroy per-bit additivity, so this is not full key recovery, and digit-count predicates don't plug directly into HNP/lattice machinery the way Minerva-style bit-length leaks do. (Fixed-point repeat submissions don't help a remote attacker: the unknown per-point timing offset is comparable to the signal and can't be calibrated without same-machine profiling.)

## Masked control: the mean channel is flat

Identical measurement against the patched call (100 secrets × 2000 samples): **corr(z_true, masked_mean) ≈ −0.10 to −0.16, consistent with 0**; per-secret means spread ~800–1400 ns, fully explained by estimation noise. The statistic the attack relies on is eliminated, exactly as the PR's expectation argument predicts.

## The covariance channel (this PR's open question), measured

`Var[R_s] = 2σ² + 2C(s) + σ²_ε` — I estimated `C(s)` directly with paired measurements of `f(t)` and `f(s+t)` (600k blocks, randomized per-block order, stall-filtered, drift-corrected block-demeaned estimator):

- Hard bound from ANOVA of unmasked `f`: `|C(s)| ≤ ~4×10⁴ ns²` for any secret.
- Measured `C(s)` for 4 secrets (incl. sparse/short canaries): values in [−8×10³, +10×10³] ns², spread ≈ **1.8×10⁴ ns²** (bootstrap 95% CI [1.6×10⁴, 2.1×10⁴]). So `Var[R_s]` is in fact weakly secret-dependent at laboratory precision — the PR's caveat that masking does not establish distribution independence is confirmed, not just hypothetical.

Exploitability of that residual:

- **End-to-end at the patched wrapper** (observed Var ≈ 1.4–2.5×10⁸ ns²): no variance difference detectable at N=20,000 per secret (permutation p = 0.36); detecting even the *maximal allowed* effect would need **~3×10¹¹ samples**. Remote HTTP noise pushes this further out.
- **Even with an idealized noiseless channel**, ~1.5×10³ samples would distinguish two *known* candidate secrets — a fingerprint, not key recovery. `C(s)` is a global autocorrelation property of the whole scalar (carries in `s+t` and the wNAF recoding couple all bits); it has no per-bit decomposition, so Kocher-style iterative bit recovery does not apply to it.

## Bottom line

The measurements confirm the PR does what it claims — the practical mean-based statistic is verifiably gone — and quantify the residual it disclaims: a real but fingerprint-grade, non-decomposable second-order signal sitting ~4 orders of magnitude below wrapper-level noise. Two caveats remain: the margin currently comes from noise rather than distribution independence (noisy-leakage scaling: attack complexity ∝ σ² for two shares), and this analysis covers only the total-timing channel, not per-iteration observation (cache/EM single-trace attacks recover each share separately; cf. the Belaïd–Rivain CHES 2023 case study). This supports the masking as a worthwhile mitigation, while the constant-time native path (`secp256k1_ecmult_const` via the ECDH API, per `security-reviews/secp256k1-timing.md`) remains the proper fix.

PR #1176, opened 2026-09-21
feat(mint): enforce keyset redemption balances

callebtc:
## Summary
- atomically cap proof reservations by the issuing keyset's recorded balance
- use checked debits for final proof invalidation and eliminate the swap/auth double-debit window
- retain cancellation-safe pending-proof cleanup

## Testing
- focused keyset balance, melt finalization, and blind-auth tests
- ruff, mypy, and pre-commit hooks


a1denvalu3 (commented):
ACK

orangeshyguy21:
This change prevents redemption above balance, but if that state is lost or damaged (via past drift in an interrupted process, or partial or complete database loss) it could prevent honest redemption.

Reconciling balance at mint startup could help with correcting past drift automatically

The lost data / disaster scenario presents challenges to the stateful guard, but something like an rpc command for manual balance overrides would be useful in those events

